PR #69, opened 2021-02-20
bugfix: reset download paths upon upgrade to default

### LINKED ISSUE #61: bug: Reset aredn.@downloads[0] values upon firmware upgrade
The URL paths in **aredn.@downloads[0]** are not being set correctly upon firmware upgrades.  Should be set to correct paths for the firmware version being installed.

dman776:
fixes #61


ae6xe (approved):
looks good.

PR #70, opened 2021-02-25
bugfix: remove freifunk feeds

ae6xe:
temporary fix until received from upstream

The feed is being removed because:
a) it is an external project and the OpenWrt team does not have access to it.
b) upon original addition of the feed, there was only a very weak tendency for the addition.
c) there is a general lack of interest in the freifunk repo to review and/or merge pull requests.
d) as far as can be found, all projects which use the freifunk feed have their own make system and self-maintained feeds list. They do not use the feeds.conf.default from the openwrt repo.

more information can be read at the following links:

http://lists.openwrt.org/pipermail/openwrt-devel/2021-February/033807.html
freifunk/openwrt-packages#37

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,7 @@ openwrt-clean: stamp-clean-openwrt-cleaned .stamp-openwrt-cleaned
 	rm -f .stamp-unpatched
 	ln -sf $(TOP_DIR)/patches $(OPENWRT_DIR)/
 	ln -sf $(TOP_DIR)/files   $(OPENWRT_DIR)/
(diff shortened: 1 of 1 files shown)


dman776 (approved):
Good solution for now.

pmelange:
https://github.com/openwrt/openwrt/commit/20caa68fec4fe033f72c9d488639f8dd2bcfa02c
https://github.com/openwrt/openwrt/commit/ff6b36b9548bbb869c2f5a5ffc0e415e25bc857e
With the above commits, the freifunk feed was removed from openWrt master and 21.02 branch.  If you are still using 19.07, then maybe you should keep this "temporary" fix :)

PR #72, opened 2021-03-04
Update alert banner background color

ab7pa:
Update the alert banner background color to distinguish it from the 
"supported hardware" messages. This makes it look more like a standard 
sticky note.


ae6xe (approved):
Looks good.

PR #73, opened 2021-03-05
Upgrade path reset2

dman776:
Tested ok via local build


dman776:
@ae6xe should be good to go

ae6xe (approved):
looks good.  Tested on a node, didn't find any issues.

PR #74, opened 2021-03-07
Clean up stale message files

ab7pa:
Clean up any stale aredn_message or local_message files at the start of 
each check.

### DIFF
--- a/files/usr/local/bin/aredn_message.sh
+++ b/files/usr/local/bin/aredn_message.sh
@@ -33,11 +33,13 @@ true <<'LICENSE'
 
 LICENSE
 
+# clean up any stale message files
+[ -f /tmp/aredn_message ] && rm /tmp/aredn_message
+[ -f /tmp/local_message ] && rm /tmp/local_message
+
 # does the node have access to downloads.arednmesh.org
 ping -q -W10 -c1 downloads.arednmesh.org > /dev/null &&
   online=true;
-  [ -f /tmp/aredn_message ] &&
-  rm /tmp/aredn_message
 
 nodename=$(echo "$HOSTNAME" | tr 'A-Z' 'a-z')
 


ae6xe (commented):
Their is an issue if a node admin connects to the internet, downloads the messages, then turns off the internet gateway.   In this use case, the message life time would only be 12 hours or less.  The message in this code is deleted on the next successive crontab run of this script if the download URL is no longer accessible.   If the message is still accessible at the URL path, it is downloaded on each run of the script.     Suggest  to put  in a test to allow a minimum age time (maybe several days) to have gone by before removing the messages to give ample time for the message to be seen in all situations.    The logic right now will keep showing the message, only if it is still downloadable.

ab7pa:
Hopefully @dman776  will weigh in, too.  I'm not really using the AAM feature for downloading messages posted on the AREDN Internet server, but have concentrated on using local_messages for non-Internet connected mesh nodes.  The existing logic removes aredn_message whenever there IS an Internet link to the AREDN server, and then re-downloads it.  But there currently is no clean-up for the local_message file.  If we don't clean them up at the beginning of the checks, then could something be added to clean-up local_message, at least?  This PR just raises the issue with a possible approach.

ab7pa:
https://www.youtube.com/watch?v=jURFNAHxBCM

ab7pa:
Local Alerts can be fairly dynamic in nature.  But here's an example issue:  A node with Local Alert messages is redeployed so the owner resets alerts[0].localpath to default.  Or worse yet [Issue #8] occurs and alerts[0].localpath gets cleared unintentionally.  Right now the node itself has no way to remove the stale local_message file, so it could persist indefinitely.  My thought is to let each node clear any stale message files on its own, and if it still _has_ Internet access or a localpath it will download and display the latest information available.  If the node _doesn't_ have Internet access or a localpath then I'm not sure of the purpose for retaining the stale message banners.

ab7pa:
I can go ahead and close this PR until the best solution is decided upon.  Many thanks.

PR #75, opened 2021-03-11
bugfix: re-submit path reset2 (previously approved)

dman776:
merge issues with prior PR.  resubmitting.


ae6xe (approved):
looks good, previously tested on prior PR.

PR #76, opened 2021-03-11
feature: add advanced config option to purge AREDN Alert msgs

dman776:
deletes AREDN and local alert messages immediately from the node


ab7pa (approved):
Tested on a node with Local Alerts feature enabled.  Worked perfectly.  
THX @dman776 dman

PR #78, opened 2021-03-15
bugfix: uci_set_indexed_option() inserting multiple empty sections

pmilazzo:
uci_set_indexed_option() inserted an empty section on every call
because uci_get_sectiontype_count() always returned 0


dman776 (approved):
Looks good.  Nice catch

PR #82, opened 2021-03-21
bugfix: Tunnel servers limited to 9 - same fix as vtundsrv in f4a1e73

pmilazzo:
Trivial fix, same as the one to vtundsrv in f4a1e73189862f6308de54deffa2862b2fafb17c


dman776 (approved):
Looks good

PR #84, opened 2021-03-26
feature: runtime-configurable tunnel limits

pmilazzo:
New Advanced Configuration entries allow changing the tunnel client and
server limits to values other than 10. If a limit is increased,
additional entries can immediately be added on the server or client
page; if reduced, entries over the limit will be retained but
deactivated upon making the configuration active. The list of tunnel
interfaces is dynamically adjusted to match the changes.

Advanced Configuration entries aredn.@tunnel[0].maxclients and
aredn.@tunnel[0].maxservers only appear when tunnel software is
installed.


dman776 (commented):
I want to take time to test this.  there is a lot of code changes/additions here.  Just want to be safe.   Let's defer this to the next release cycle.   @ae6xe what do you think?

ab7pa:
I did some testing only on the _Tunnel Client_ side.  When I set MaxClients = 1 the UI allows me to add only a single client entry (as I would expect).  If I change MaxClients = 2 and go back to the TunClient UI, I'm still only allowed a single client entry -- no blank row with ADD button.  I haven't had a chance to test MaxClients > 10 yet.  I'll also try to do some testing on the _Tunnel Server_ side in the near future.

pmilazzo:
Steve:

Thank you very much for testing this feature!

I think there is some potential confusion here with the words "server" 
and "client":

  * The "Tunnel Client" setup tab is a list of /servers /to which this
    tunnel client should connect (the text above the list reads,
    "Connect this node to the following servers:"),//and thus is
    controlled by the /maxservers/ variable.
  * Conversely, the "Tunnel Server" setup tab is a list of /clients
    /that may connect to this tunnel server, and is thus controlled by
    the /maxclients /variable.

I repeated your test, except I changed maxservers rather than 
maxclients. After setting maxservers to 2, the "Add" button reappeared 
on the "Tunnel Client" tab, as it should.

So, the nomenclature is confusing, and I'm not sure what to do about it. 
The help message for ***@***.***[0].maxservers currently reads,

    Specifies the maximum number of tunnel servers to which this node
    can connect.  (Only applies if tunnel software is installed)

...but that's clearly not enough help. And you have to hover over the 
"?" before you see it. So that's not happening either. Suggestions?

73 and thanks again,

- Paul, K3PGM

ab7pa:
Yes, it's easy to confuse them.  I misspoke in my first comment.  When I set MaxServers = 1 the TunClient UI allows me to add only a single client entry (as I would expect). If I change MaxServers = 2 and go back to the TunClient UI, I see the blank row with ADD button.  After adding the 2nd server credentials row, I change MaxServers = 3 and I again see a blank row with ADD button.  With 2 server credential rows entered, if I change MaxServers back to "1" I still see the 2nd set of credentials (two rows show up).

I'm arranging for more external client and server credentials so I can continue testing.  THX

pmilazzo:
> With 2 server credential rows entered, if I change MaxServers back to "1" I still see the 2nd set of credentials (two rows show up).

Right, because I didn't want to delete anyone's credentials. After you press "Save Changes", though, only one tunnel will come up: the first one with a check mark in the "Enabled?" box. That's the way the software has always worked, and I didn't change that behavior, but it was less obvious before the number of servers was adjustable.

One could argue that the UI should only allow you to check as many boxes as you have tunnels configured, or, failing that, at least clear the check boxes of the tunnels that exceed the current limit. I should probably figure out how to do one of those...

> I'm arranging for more external client and server credentials so I can continue testing. THX

That's great, thanks! I really appreciate your help testing this stuff.

Thanks and 73,
- Paul, K3PGM

ab7pa:
Right now the advancedconfig display has two new rows for this PR:
aredn.@tunnel[0].maxclients & aredn.@tunnel[0].maxservers 
The hover help is good, but the key names could be confusing.
Would it clarify things to use two different keys, maybe something like this?
aredn.@tunclient[0].maxservers & aredn.@tunserver[0].maxclients
Just thinking out loud.

dman776:
Yeah the names are wierd.   I struggled with them when I wrote originally.   Wonder if we just change to “Tunnel Inbound” and “tunnel outbound” settings

Thanks,
Darryl

ab7pa:
I was able to get another set of tunnel client credentials.  I increased maxservers and got the new row with ADD button to enter the creds.  The tunnel came up fine.  Went back and decreased maxservers below the number of active tunnels I currently have running.  All active tunnels still active even with lower setting, but when I click _Save Changes_ on the TunClient UI the bottom active tunnel in my list goes offline (which is exactly how you described it should work).  In all my testing of the Tunnel Client settings so far, everything works exactly as it should.  

I'm going to get permission to test your Tunnel Server settings on a local tunsvr which has a full 10 clients and wants to add more.  I'll let you know how that goes as I test the tunnel server side of the code.  Many thanks.

ab7pa:
So far on the Tunnel Server side I've only tested the TunServer UI and it functions as intended.  I only tested up to 15 "dummy" tunnel clients, but everything works as designed.  

This test server was also running 2 tunnel client connections, and when I look at the interfaces I see tun60 and tun61 (the two client interfaces).  In the past, tun50 - tun59 were the designated tunnel server interface names, and tun60 - tun69 were designated as client interfaces. I have not been able to verify whether the new tunnel server interface identifiers will overlap with the existing client range.  If I have 15 or 20 clients on this tunnel server, what interface identifiers will be used for them?

pmilazzo:
> This test server was also running 2 tunnel client connections, and 
> when I look at the interfaces I see tun60 and tun61 (the two client 
> interfaces). In the past, tun50 - tun59 were the designated tunnel 
> server interface names, and tun60 - tun69 were designated as client 
> interfaces. I have not been able to verify whether the new tunnel 
> server interface identifiers will overlap with the existing client 
> range. If I have 15 or 20 clients on this tunnel server, what 
> interface identifiers will be used for them?

The ranges of interface numbers will change as you change the maximum 
number of clients and servers, but will remain contiguous. The 
interfaces listed in the configuration files will change as soon as you 
save the changes in the Advanced Configuration, but the actual 
allocation of interfaces to tunnels will only change when you either 
reboot or press "Save Changes" in one of the two Tunnel tabs. So:

  * In the default configuration, you start with tun50 -- tun69
    configured, with server interfaces tun50 -- tun59 and client
    interfaces tun60 -- tun69.
  * If you now set maxclients=15, five more interfaces will be
    configured, so now you have tun50 -- tun74, but live tunnels will
    not move, so your two client endpoints are still tun60 and tun61.
  * Now you either reboot or press one of the "Save Changes" buttons.
    The two client endpoints will move to tun65 and tun66, making tun50
    -- tun64 available for server endpoints.
  * If you now set maxclients=5 and maxservers=5, the list of configured
    tunnels will shrink to tun50 -- tun59, but the two clients endpoints
    will remain live at tun65 and tun66 because only the configuration
    files have changed, not what's going on in the kernel.
  * After the next reboot or Save Changes, the two client endpoints will
    have moved to tun55 and tun56.

- Paul, K3PGM

ab7pa:
Thanks, Paul.  That makes sense to me, and it seems to work as designed in all of the testing I've done.  Seems good to go as far as I'm concerned.  Many thanks!
Steve  ab7pa

pmilazzo:
Steve:

Thank you again for digging in here; I really appreciate it!

I'm eager to hear the results on the overstuffed tunnel server you 
mentioned earlier. Just make sure they have enough RAM!

FYI, I'm using a GL.iNet GL-AR750 as a big tunnel server; 128 MB RAM for 
$40! Our tunnels aren't terribly busy, though; I'm not sure how well it 
would do with, say, a dozen simultaneous HD video streams...

73,

- Paul, K3PGM

ab7pa:
I set up a tunnel server node with the new code and have been slowly adding client connections as others here opt in to participate.  So far the new tunnel server is running 12 clients and is barely "breaking a sweat."  CPU ~ 80-85% idle and Memory Usage ~ 32mb on a 64mb node.  Each vtund process has a VSZ footprint of exactly 1256kb.   I will keep adding clients as others opt in, and I'll report back on node resource usage.  So far I'm impressed with the performance.

ab7pa:
I ran the "stress test" over the weekend and the code behaved very well.  I ramped up the number of connections and the UI handled everything as expected.  I also set the max to zero and verified that the UI does not allow adds with that setting.  As far as memory usage, the node only allocates memory when the connection becomes active so there can be more clients in the UI than memory allocated if some are inactive.  I would recommend that we merge this PR so it can be included in an upcoming Nightly Build, which would allow many others to test it.  Many thanks, Paul.

dman776:
Good point.   We should not allow values <= 0

Thanks,
Darryl

ab7pa:
Negative integers would definitely not make sense.  Zero values could be used to disable one or the other types of tunneling.  For example, some people running a tunnel server might want to disable using it as a tunnel client node, and some who run a node for tunnel client connections might want to prevent it being used as a tunnel server.

pmilazzo:
I, too, consider 0 a legitimate value, indicating that you want to 
devote all your resources to tunnels in the other direction. I tested 
this case, and it appears to behave properly.

Negative numbers, on the other hand, are definitely a problem and should 
be prohibited. There should probably also be an upper limit. but I'm not 
sure where to set it; tun99999999 is a perfectly valid interface name, 
but impractical. An AR-GL150 has enough RAM for at least 50 active 
tunnels; some of the new 802.11ax platforms have 512 MB, enough for 
several hundred active tunnels (at about 3 bit/sec each, but active 
nonetheless). And inactive tunnels consume few resources.

So what's a good upper bound, reasonably future-proof but not silly? 50? 
100? 500?

I'm happy to add range-checking to the software, but I'll need a few 
hints on the AREDN-approved way to deal with invalid input in the UI. 
While I'm at it, I should also probably limit the number of tunnels that 
can be marked "Enabled" at once; is there a callback somewhere I could 
intercept?

Thanks and 73,

- Paul, K3PGM

> Negative integers would definitely not make sense. Zero values could 
> be used to disable one or the other types of tunneling. For example, 
> some people running a tunnel server might want to disable using it as 
> a tunnel client node, and some who run a node for tunnel client 
> connections might want to prevent it being used as a tunnel server.
>

ab7pa:
Paul is adding some data validation now, too.  THX

pmilazzo:
I've now added two types of data validation:

- Commit #69f4953 causes the Tunnel Server and Tunnel Client tabs to display an error if the number of tunnels that are marked "Enabled" exceeds the current value of maxclients or maxservers, respectively. The user can clear the error either by increasing the limit or by unchecking enough "Enabled" boxes to reduce the number of enabled tunnels to the limit.
- Commit #f45547a implements range checking for the maxclients and maxservers settings on the Advanced Configuration tab. The values must now be integers in the range of [0,$tunnelLimitsUpperBound], where $tunnelLimitsUpperBound is currently set to 100 in the code. Integer values that exceed these limits are clamped to the limit; values that are not integers are treated as "0".

73,
Paul, K3PGM

ab7pa:
Great work, Paul!

dman776 (approved):
Nice work Paul

ravetildon:
What device or devices were used in the testing that didn't break a sweat?

PR #85, opened 2021-03-26
Update perlfunc.pm

kc8ufv:
Added new hardware model strings for the Mikrotik Routerboard 911 & 912, since Mikrotik now puts them as RB911/RB912


ae6xe (approved):
Looks good.   The entries in openwrt/target/linux/ar71xx/base-files/lib/ar71xx.sh already handle these models, so good there too.

PR #88, opened 2021-04-06
bug: fix nill typo

ab7pa:
Fix typo (nill to nil) in pidfile test.

### DIFF
--- a/files/usr/local/bin/snrlog
+++ b/files/usr/local/bin/snrlog
@@ -282,7 +282,7 @@ local f = io.open(pidfile,"r")
 if (f) then
   local oldpid = f:read("*number")
   f:close()
-  if (oldpid ~= nill and dir_exists("/proc/" .. oldpid)) then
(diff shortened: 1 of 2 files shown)


ae6xe (approved):
good catch.

ae6xe:
did this fix the spike to 0dBm issue?  What symptoms will go away?

ab7pa:
No, this was totally unrelated to the "spike to zero" issue.  Just happened to catch this _nil_ issue while looking through those two files.

ae6xe:
My only concern is that this would change the behavior in some way.   We presume this would be good.  But possibly, it is bad as this code was ether always run or never run.     @dman776, since you wrote this code originally, would you do a quick review to confirm we should commit this.

ab7pa:
I agree.  Seems like the first part of this test will always pass, since _oldpid_ will never equal _nill_

if (oldpid ~= nill and dir_exists("/proc/" .. oldpid))

dman776 (approved):
good catch

PR #89, opened 2021-04-06
Snr zero spike fix

ab7pa:
Fix the SNR spike to zero on the Charts.  This sets the value to the default floor.

### DIFF
--- a/files/usr/local/bin/snrlog
+++ b/files/usr/local/bin/snrlog
@@ -282,7 +282,7 @@ local f = io.open(pidfile,"r")
 if (f) then
   local oldpid = f:read("*number")
   f:close()
-  if (oldpid ~= nill and dir_exists("/proc/" .. oldpid)) then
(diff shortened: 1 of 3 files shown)


ab7pa:
Chart values before fix:
existing-API-value-without-fix

Chart values after fix:
new-Chart-values-with-fix

ab7pa:
### Problem:  I just noticed that I build PR#89 on top of my previous PR#88 branch.  UGH!  Sorry about that.  PR#89 contains both sets of fixes.  Let me know if I need to kill PR#89 and revert everything to main before adding a separate branch for the SNR zero spike fixes.  So sorry about that.  Never occurred to me until I’d already submitted it.

ab7pa:
Closing this PR and resubmitting another one that is not built on a previous PR's branch.  My mistake.  I'll correct this and resubmit.  Thanks.

PR #90, opened 2021-04-06
bug: Fix spike to zero in SNR Chart

ab7pa:
Fix spike to zero in SNR Chart.

Chart before fix:
1-existing-API-value-without-fix

Chart after fix:
2-new-Chart-values-with-fix


dman776 (approved):
looks good.   sending null values to -95

PR #94, opened 2021-04-21
API Additions

dman776:
feature: additional api functions: api_version, totalram, totalfreespace, alerts

PR #95, opened 2021-04-21
Lua info gateway

dman776:
correct bug in info.lua library when there is no internet gateway accessible


dman776:
@ab7pa didnt you test this?

ab7pa (approved):
@dman776  I tested this and it solved the issue I was having with an api error.  Seems like this PR includes everything from #94 plus the fix in getDefaultGW().  Is #94 needed if everything is in #95?

PR #99, opened 2021-05-05
bugfix: fix api when meshrf is disabled

### LINKED ISSUE #98: bug: api should handle nil device if meshrf is disabled
The _getStatusMeshRF_ function in /www/cgi-bin/api expects a valid "device" to be returned in order to get the channel, bw, band, and freq.  If the node has MeshRF disabled, no valid "meshRadioDevice" will be returned.  Code should handle a nil "device" -- this will be required in order for the newui to populate all the node values.

function getStatusMeshRF()
  local info={}
  info['ssid']=aredn_info.getSSID()
  info['device']=aredn_info.getMeshRadioDevice()
  **if info['device'] ~= "" then**
    info['channel']=aredn_info.getChannel(info['**device**'])
    info['chanbw']=aredn_info.getChannelBW(info['**device**'])
    info['band']=aredn_info.getBand(info['**device**'])
    info['frequency']=aredn_info.getFreq(info['**device**'])
  **end**
  return info
end

dman776:
fixes #98

### DIFF
--- a/files/www/cgi-bin/api
+++ b/files/www/cgi-bin/api
@@ -76,12 +76,15 @@ end
 
 function getStatusMeshRF()
 	local info={}
-	info['ssid']=aredn_info.getSSID()
-	info['device']=aredn_info.getMeshRadioDevice()
-	info['channel']=aredn_info.getChannel(info['device'])
-	info['chanbw']=aredn_info.getChannelBW(info['device'])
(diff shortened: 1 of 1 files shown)


ab7pa (approved):
Tested on node with **meshrf** _enabled_ and _disabled_.  Works perfectly.  Thanks, @dman776

PR #100, opened 2021-05-05
Format Uptime as String

ab7pa:
Format uptime as a string with values for days, hours, minutes, and seconds.
Added this to /usr/lib/lua/aredn/info.lua in the model.getUptime function.

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -374,15 +374,9 @@ end
 -- Current System Uptime
 -------------------------------------
 function model.getUptime()
-        local mynix=nixio.sysinfo()
-        local upsecs=mynix['uptime']
-        local days=math.floor(upsecs / 86400)
-        upsecs=upsecs - days * 86400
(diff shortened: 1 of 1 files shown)


ab7pa:
I just found there is a secondsToClock function in utils.lua which does the same thing.  I'm going to take advantage of that function rather than reproduce the code in info.lua

ab7pa:
Okay, the secondsToClock function does what is needed.  THX

dman776:
In this case, it's intentional.  The presentation should format into the locale specific format.  Ie.  If my browser locale is US English, it should format as days hr:man:sec, but other locales may be different.

ab7pa:
Okay, thanks @dman776 .  In the newui I worked with today it shows up simply as a number of seconds (upsecs) without any formatting, so it looked like it needed some help.  If I'm understanding you correctly, I should close this PR as unnecessary.  Much appreciated.

dman776 (approved):
actually, after re-consideration, this is NOT a "time" value.   it is elapsed time.   i think it ok to put this conversion here instead of the frontend.

PR #101, opened 2021-05-06
api enhancement: convert getFreq to use lib-iwinfo

dman776:
remove "GHz" label
returns as MHz value (ie. 2402)


r1de (approved):

r1de (commented):
yes. much better!
I will probably have to change the MeshMap but it's just a cosmetic thing :crossed_fingers:

PR #102, opened 2021-05-08
fix: Local Alert File Read All

ab7pa:
Read the entire local_message file rather than just the first line.


dman776 (approved):
good catch.

ab7pa:
The newui has a _v-show_ test for displaying the alert button:  `alert.aredn !== ' '`
Adding the final _if-else_ in the getAlert functions will ensure that test isn't confused by an empty alert file.

PR #106, opened 2021-05-20
Remove local.mesh suffix from CurrentNeighbor hostname

ab7pa:
In olsr.lua there are gsubs to remove dtdlink and mid prefixes.  Add a gsub to remove the local.mesh suffix from hostname too.


ab7pa:
Looking at the newui Local Hosts page I see all the hostnames are lowercase.  I think we should make them consistent -- either all uppercase or all lowercase.  I'm changing this PR so it does a string.lower(host) to make it consistent with the Local Hosts page.

dman776 (approved):
i'm ok with this since it is in the api.   All ACTUAL links should always include .local.mesh as they are created in the frontend.

PR #107, opened 2021-05-20
Remove local.mesh suffix from Local Hosts for newui

ab7pa:
This removes the local.mesh suffix from hostnames returned through the **api** for the Local Hosts display in the newui.  Similar change as was submitted for the Current Neighbors display earlier.


ab7pa:
Closing this PR due to issues with the commits.  Will open a new PR that includes ONLY the change intended.

PR #108, opened 2021-05-20
Remove local.mesh suffix from Local Hosts hostname in newui

ab7pa:
Remove the local.mesh suffix from hostnames returned through the **api** for the newui Local Hosts display.


dman776 (approved):
looks good

PR #109, opened 2021-05-21
Add LocalHost Connection Type

ab7pa:
Add a connection type field to **info.lua** so the **api** can provide it to the newui _Local Hosts_ display.


dman776 (approved):
nice addition

PR #110, opened 2021-05-22
Fix regex for current neighbor midxx.hostname

ab7pa:
Fix regex for mid%d.hostname. It's possible that there may be more than a single digit. This PR also puts all the string.gsub processing in the if-else for _host_ rather than _remhost_.


dman776 (approved):
Good catch

PR #111, opened 2021-05-23
Add Default Value for Description

ab7pa:
Add a default value of _None_ if no description exists.
This could be handled in the newui, but if no description exists then the description key will never appear in the data returned by the _api_.  This way, something is always returned and the key will always appear in the data.


dman776:
this should be left to the ui to handle a "" value for description.  closing.

PR #112, opened 2021-05-24
Api use libiwinfo frequency

dman776:
add support for estimated_throughput to api

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -229,11 +229,10 @@ end
 -------------------------------------
 function model.getFreq()
 	local wlanInf=get_ifname('wifi')
-	local freq=""
-	freq=os.capture("iwinfo " .. wlanInf .. " info | egrep 'Mode:'")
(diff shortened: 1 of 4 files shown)


ab7pa (commented):
Do we need to standardize on uppercase vs. lowercase for all host/hostname values and include that in **info.lua** and **olsr.lua**?  Currently we have a mixed case situation.  This decision may ripple across any _string.match_ code throughout.  Seems like now would be a good time to standardize on one case or the other.

If you decide to eliminate the _nslookup_ calls, that will help too.  _nslookup_ can return mixed case values like this:  **mid4.KB7RJ-hAPac**

ae6xe:
suggest to do the same as nslookup and return the same upper/lower case hostname combination.   Preserve the case to always display for end user readability.   Hostnames can be more readable, e.g.   AE6XE-MountainNode .     Since hostnames are case insensitive (hosts XYZ and XyZ would be in conflict), any comparisons or matching needs to also be case insensitive at that point.

ab7pa (approved):
Code works as designed.  **api** shows correct throughput estimate for RF interfaces, as well as entries for services for all nodes.

PR #114, opened 2021-05-26
add traceroute and remotenodes to api

dman776:
traceroute, remotenodes to api

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -288,6 +288,11 @@ function model.all_services()
 				service['link']=link
 				service['protocol']=protocol
 				service['name']=name
+				if ip==" my own service" then
+					service['ip']=model.getInterfaceIPAddress("wifi")
+				else
+					service['ip']=ip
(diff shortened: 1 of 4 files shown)


r1de (approved):
looks ~~ok~~ pretty awesome to me.
That'll be useful!
If you didn't get a "Bad Gateway" when testing it, then it is probably just fine!

ab7pa (changes requested):
I tested this PR on top of all previous ones in the queue (PR112 & PR113) and it worked well, with a couple of tweaks.  PR113 removes hostname case manipulation, so this PR114 should not uppercase the hostname (info.lua lines 331 and 551).  Also utils.lua should probably use` %d+` on line 277 to catch one or _more_ digits in the **mid** prefix.  Then in api the `name:match` should have lowercase _mid_ on line 135.

ab7pa (approved):
I tested this PR on top of NB368 today and it worked well.

ab7pa:
Is there more to do on this PR or is it ready to merge?  Looks fine to me.

PR #115, opened 2021-05-28
add invalid link handling to api

ab7pa:


dman776 (approved):
i wish there was a better place to do this.  but, it should be ok.

PR #119, opened 2021-06-06
leave mixed hostname case in api

ab7pa:
Leave hostnames as natural mixed case.

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -327,8 +327,8 @@ function model.all_hosts()
 				local ip, name=string.match(data,"^([%x%.%:]+)%s+(%S.*)\t%s*$")
 				if ip and name then
 					if not string.match(name,"^(dtdlink[.]).*") then
-						if not string.match(name,"^(mid[0-9][.]).*") then
-							host['name']=name:upper()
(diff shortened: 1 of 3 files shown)


ab7pa:
There may still be something slipping through the cracks because I'm not seeing the **expected_throughput** numbers returned by the **api**.  All other values are coming through okay.

ab7pa:
After further testing ... the reason the **api** did not return the 5 _stnInfo_ metrics was due to the state of OLSR on my node at the time.  The mac2node() function parses items in /tmp/snrlog -- and one of my entries there had the mac prefix, the dash character, but the nodename/ipaddr were missing.  Nothing after the dash.  After restarting OLSR and waiting an appropriate time for OLSR to populate everything, then mac2node() returned the nodename/ipaddr and the **api** returned the 5 _stnInfo_ metrics.  Seems like we may need to add some error checking in case /tmp/snrlog entries get into an odd state.

ab7pa:
As long as the mac2host() function returns the expected values, the code in this PR behaves as expected.

ab7pa:
Added a line that handles the rare case when the nodename or ip address is missing from the snrlog filename.  Sets node to remnode in this rare instance.

ab7pa:
Closing this PR since case manipulation is handled in aredn/aredn PR#124

PR #120, opened 2021-06-12
api: add getRemRFIP function

ab7pa:
Add a getRemRFIP() function to return the RF IP address even for non-RF linked nodes. This allows services to be displayed for those nodes on the Current Neighbors tab.  This is one proposed solution for missing services on tunnel connected nodes (for example).

### DIFF
--- a/files/usr/lib/lua/aredn/olsr.lua
+++ b/files/usr/lib/lua/aredn/olsr.lua
@@ -51,6 +51,14 @@ function model.getOLSRRoutes()
   return routes['routes']
 end
 
+-- Get the RF IP address of a neighbor node
+function model.getRemRFIP(ip)
+  local url=string.format("http://%s/cgi-bin/api?status=ip", ip )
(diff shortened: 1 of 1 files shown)


files/usr/lib/lua/aredn/olsr.lua
```diff
@@ -51,6 +51,14 @@ function model.getOLSRRoutes()
   return routes['routes']
 end
 
+-- Get the RF IP address of a neighbor node
+function model.getRemRFIP(ip)
+  local url=string.format("http://%s/cgi-bin/api?status=ip", ip )
```
dman776, line 56: I think this is a no-no.   We should have this info from OLSR locally on the node instead of generating more network traffic.

dman776 (changes requested):

ab7pa:
Cool.  I'll close this approach and we can come up with a better solution.  THX

PR #124, opened 2021-06-17
api: add RF IP for non-RF links

ab7pa:
Add the RF IP address for non-RF linked devices. This allows the services to be displayed for non-RF Current Neighbors.


dman776 (approved):
looks good

dman776 (approved):
looks good

PR #125, opened 2021-06-18
aredn: added network topology to mesh info API

r1de:
~~Displays~~ Returns the network topology in a nice json format.
Shows much more information than the "dotdraw" plugin and is in a different format than the "dotdraw" plugin.


r1de:
I could move the `fetch_json("http://127.0.0.1:9090/topology")` call to `olsr.lua`.
I just thought it might be less overhead by not having to call yet another another function to just do the same thing.

dman776 (approved):

PR #126, opened 2021-06-18
Link info changes

r1de:
changes the sysinfo_json?link_info=1 information to use the actual mesh IP of the DTD and Tunnel linked nodes rather than the internal routing IP's used before.


r1de:
I may just not be fully awake yet, but I do not understand the conflict...   :facepalm:

r1de:
re-creating....

PR #127, opened 2021-06-19
aredn: changes for sysinfo.json link_info from info.lua

r1de:
changes the sysinfo_json?link_info=1 information to use the actual mesh IP of the DTD and Tunnel linked nodes rather than the internal routing IP's used before.
More consistant with topology output about linked nodes.

### DIFF
--- a/files/usr/lib/lua/aredn/utils.lua
+++ b/files/usr/lib/lua/aredn/utils.lua
@@ -258,6 +258,20 @@ function nslookup(ip)
 end
 
 -------------------------------------
+-- Returns first IP of given host
+-------------------------------------
+function iplookup(host)
+	local ip=nil
+	local nso=nil
+	if host:find("dtd.*%.") or host:find("mid%d%.") then
+		shortname=host:match("%.(.*)")
+	end
+	nso=capture("nslookup "..host)
+	ip=nso:match("Address 1: (.*)%c")
+	return ip
+end
+
+-------------------------------------
 -- Returns traceroute
 -------------------------------------
 function getTraceroute(target)
--- a/files/usr/lib/lua/aredn/utils.lua
+++ b/files/usr/lib/lua/aredn/utils.lua
@@ -264,7 +264,7 @@ function iplookup(host)
 	local ip=nil
 	local nso=nil
 	if host:find("dtd.*%.") or host:find("mid%d%.") then
-		shortname=host:match("%.(.*)")
+		host=host:match("%.(.*)")
 	end
 	nso=capture("nslookup "..host)
 	ip=nso:match("Address 1: (.*)%c")
--- a/files/usr/lib/lua/aredn/olsr.lua
+++ b/files/usr/lib/lua/aredn/olsr.lua
@@ -71,13 +71,15 @@ function model.getCurrentNeighbors(RFinfo)
     local host
     local linkip=v['remoteIP']
 
+    local linkhost=nslookup(linkip) -- TOTO: stop using nslookup? use /var/run/olsr_hosts
+    linkip=iplookup(linkhost)
+
     info[linkip]={}
     info[linkip]['olsrInterface']=v['olsrInterface']
     info[linkip]['linkType']= model.getOLSRInterfaceType(v['olsrInterface'])  -- RF or DTD or TUN
     info[linkip]['linkQuality']=v['linkQuality']
     info[linkip]['neighborLinkQuality']=v['neighborLinkQuality']
 
-    local linkhost=nslookup(linkip) -- TOTO: stop using nslookup? use /var/run/olsr_hosts
     if linkhost~=nil then
       host = string.gsub(linkhost,"mid%d+.", "")
       host = string.gsub(host,"dtdlink%.", "")
--- a/files/usr/lib/lua/aredn/olsr.lua
+++ b/files/usr/lib/lua/aredn/olsr.lua
@@ -66,63 +66,73 @@ end
 function model.getCurrentNeighbors(RFinfo)
   local RFinfo = RFinfo or false
   local info={}
-  local links=model.getOLSRLinks()  -- Get info for all current neighbors
+  local links=model.getOLSRLinks() -- Get info for all current neighbors
   for k,v in pairs(links) do
-    local host
-    local linkip=v['remoteIP']
+    local host=nslookup(v['remoteIP'])
+    local mainip=iplookup(host)
+    info[mainip]={}
 
-    local linkhost=nslookup(linkip) -- TOTO: stop using nslookup? use /var/run/olsr_hosts
-    linkip=iplookup(linkhost)
-
-    info[linkip]={}
-    info[linkip]['olsrInterface']=v['olsrInterface']
-    info[linkip]['linkType']= model.getOLSRInterfaceType(v['olsrInterface'])  -- RF or DTD or TUN
-    info[linkip]['linkQuality']=v['linkQuality']
-    info[linkip]['neighborLinkQuality']=v['neighborLinkQuality']
-
-    if linkhost~=nil then
-      host = string.gsub(linkhost,"mid%d+.", "")
+    if host~=nil then
+      host = string.gsub(host,"mid%d+.", "")
       host = string.gsub(host,"dtdlink%.", "")
       host = string.gsub(host,".local.mesh$","")
-      info[linkip]['hostname']=host
+      info[mainip]['hostname']=host
     else
-      info[linkip]['hostname']=linkip
+      info[mainip]['hostname']=mainip
     end
 
-    if info[linkip]['linkType'] == "RF" and RFinfo then  -- get additional info for RF link
-      require("iwinfo")
-      info[linkip]["rfip"] = linkip
+    info[mainip]['olsrInterface']=v['olsrInterface']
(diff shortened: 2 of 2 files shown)


r1de:
is using grep (or something) more efficient that using nslookup in this instance?
I think that offloading the request to DNS (which already has all that info in RAM) is better than pulling the, possibly huge, hosts_olsr file just to look for one IP.
Could be just me tho. :grey_question:

ab7pa:
I changed the variable name to "linkip" is so it would retain the IP address of the actual link as well as adding a new value for "rfip" to hold the Mesh RF IP address that's required in order to get the services on those nodes.  In the code the "linkip" is already being set by the getOLSRlinks() function on line 72, so I'm not sure we need to set it again on line 75.  Then further down in the code the "rfip" is being set separately for all non-RF connected devices.

r1de:
Ok @ab7pa I will look and see if you already implemented what I added.
IIRC tho, getOLSRLinks() was what was returning the funky IP addresses.
I wanted to change them out of consistency and because the MeshMap "version 2" will be using a lot more info to show the links, it's easier to correlate all the info if the "real" mesh IP's are used in the `link_info` section.
That's all. :)

I'll look again tho. :+1:

r1de:
It's not the same.
The `link_info` array index numbers are pulled from the `remoteIP` values from `getOLSRLinks()`, they use the localnodes "internal" IP's.
Screenshot_2021-06-18_21-34-45

That is what I was trying to change, I think for consistency and other things those IP's should be the real IP's of the linked nodes.

ab7pa:
@r1de , I originally thought about taking the same approach you're thinking of doing.  It replaces the actual "link" IP address with the wifi/RF IP address.  I backed off that approach because you lose the actual "link" IP info in the process.  I figured there might be some benefit for retaining both IP addresses, which is what the latest code does.

I was actually thinking about changing the getCurrentNeighbors() function so it is NOT lndexed by IP address.  Neither of the other functions (getLocalHosts or getRemoteNodes) are indexed by IP.  If we did that, then both IP addresses could be retained as values in the table.  I guess we need to figure out whether there is a good reason to keep both IPs, or if the actual link IP address being used by the node is unimportant so it could be thrown away by the api.

ae6xe:
A node has several IP addresses, one for each interface.    However, only the main IP address will resolve back to the known hostname.  All the other IP addresses will resolve back to odd-ball hostnames, e.g. dtdlink.host.local.mesh .    It would probably simply things to always figure out or resolve back to the main IP address and known hostname, when referring to a node.  Then comparing or otherwise knowing what node we are talking about does not have confusion.  In the current design, a node must have a main IP address and hostname, which is derived from the MAC address of the main RF interface.

ab7pa:
If there's no benefit for keeping the IP address the node is actually using for the link, then there's no problem with throwing it away in the api.  I don't have a preference.  Whichever is most clear and simple and efficient from a resource standpoint.  Keeping the link IP allows us to show it on the Current Neighbors display -- services are pulled using the "rfip" but the actual link IP is still displayed.  If we don't want the actual link IP to be displayed to users, then it won't matter if we throw it away in the api.

ae6xe:
I'd propose if an IP address is shown, only and always show the main IP for a node to simplify for end users.  Otherwise, many different IP addresses are showing to different people from different views on the mesh, might lead to confusion for those not network savvy.    There can be numerous IP addresses with multiple tunnels, and future use of dual band devices:

KI7LXY-QRT5.local.mesh   10.1.2.3
mid2.KI7LXY-QRT5.local.mesh 172.31.29.50
dtdlink.KI7LXY-QRT5.local.mesh 10.2.2.3
mid1.KI7LXY-QRT5.local.mesh 10.1.2.3
...

One option might be a drill down -- select a node or hostname with some sort of a "network info" option that shows all the interfaces, hostnames, IP addresses, and network config details of each interface (and annotate which interface you would be using to route to this node).   Maybe even a "route path" option to show the hops to that service or node.   If we put some or all of this networking and routing details in the neighbor or remote nodes page, it risks becoming too cluttered or overly complex.   The primary focus is on the nodes we can talk to, the services we can reach,  and expected performance to communicate with that node/service.

r1de:
Simpler is better.
Most users don't need to know about the internal address(es), even us network nerds hardly even need to know about them and when we _do_, we know where to find that information.
I don't remember why the array index used the IP addresses, probably because that's just how OLSR gave _us_ the data.

I think it's easier to "mesh" all of the network data together (pun intended. :smile: ), like for my mapping system, if the indexes are left as IP's.

Just my $0.02.

ab7pa:
How about this for the getCurrentNeighbors() function?  It uses Eric's new iplookup() to get the main IP of the node.

[olsr.lua.txt](https://github.com/aredn/aredn/files/6681341/olsr.lua.txt)

files/usr/lib/lua/aredn/utils.lua
```diff
@@ -257,6 +257,20 @@ function nslookup(ip)
...
 end
 
+-------------------------------------
+-- Returns first IP of given host
+-------------------------------------
+function iplookup(host)
+	local ip=nil
+	local nso=nil
+	if host:find("dtd.*%.") or host:find("mid%d%.") then
+		shortname=host:match("%.(.*)")
```
ab7pa, line 267: Is the _shortname_ variable used somewhere else?  Or should it be **host=host:match("%.(.*)")** ?

r1de: Good catch!
I looked at it so much I missed that! :facepalm:

r1de:
> How about this for the getCurrentNeighbors() function? It uses Eric's new iplookup() to get the main IP of the node.
> 
> [olsr.lua.txt](https://github.com/aredn/aredn/files/6681341/olsr.lua.txt)

I think that looks just fine.
There is even more information that can be added to `link_info` for each link eventually.
I am starting to try and track all this stuff, it is way too much info for most purposes, but some of us think we might be able to better track down network issues with it.

ab7pa:
Okay.  Feel free to grab that getCurrentNeighbors() function and pop it into your PR if you want to.
I'm tweaking the newui code for Current Neighbors so it displays the correct info provided by that function.  THX

r1de:
roger.
give me a few minutes.

ab7pa:
Cool.  I tested it on my end and the **api** now has the _Main IP_ as the index.  This simplifies a lot of the code in that function too.

r1de:
> Cool. I tested it on my end and the **api** now has the _Main IP_ as the index. This simplifies a lot of the code in that function too.
:+1: 
That was 1/2 my point by changing it, laziness! :smile:. It was easier to deal with other code if getCurrentNeighbors() used the real IP's.
While I am updating it with your changes, I'll see if I can get this other info in there too...

r1de:
Couldn't have just left well enough alone could ya? :smile: 
Check this!
I might have missed a "mainip" vs. "linkip" vs. "remip" somewhere. :confounded: 

tested on my node and it works, but still.... :smile:

r1de:
additional info looks like this (for a DTD link)
Screenshot_2021-06-19_16-02-11

r1de:
wait. might not be working right with RF links...
I am not seeing the rx_rate and tx_rate nor the bandwidth info.... :cry:

ab7pa:
Hmmm... I'm seeing all the correct info here...  but I'm running my shorter version without all the extra values.  Here's what the api spits out for me on an RF node.

         "10.3.168.137": {
           "neighborLinkQuality": 1,
           "noise": -95,
           "linkQuality": 0.88600000000000001,
           "tx_rate": 14.4,
           "linkType": "RF",
           "hostname": "AB7PA-A15",
           "expected_throughput": 6.2999999999999998,
           "olsrInterface": "wlan0",
           "rx_rate": 16.199999999999999,
           "signal": -59
         },

ab7pa:
When I paste your additional values into my function this is what I get.  Seems correct to me.

         "10.3.168.137": {
           "helloTime": 0,
           "lostLinkTime": 0,
           "linkQuality": 0.88600000000000001,
           "vtime": 20000,
           "linkCost": 1.1279300000000001,
           "linkType": "RF",
           "hostname": "AB7PA-A15",
           "validityTime": 39064,
           "previousLinkStatus": "SYMMETRIC",
           "currentLinkStatus": "SYMMETRIC",
           "rx_rate": 14.4,
           "neighborLinkQuality": 1,
           "noise": -95,
           "symmetryTime": 19296,
           "seqnoValid": false,
           "lossMultiplier": 65536,
           "pending": false,
           "lossHelloInterval": 2000,
           "asymmetryTime": 97356267,
           "tx_rate": 14.4,
           "hysteresis": 0,
           "seqno": 0,
           "lossTime": 2296,
           "expected_throughput": 7.4000000000000004,
           "olsrInterface": "wlan0",
           "lastHelloTime": 0,
           "signal": -58
         },

ab7pa:
Here's my getCurrentNeighbors function...

[getCurrentNeighbors-function.txt](https://github.com/aredn/aredn/files/6681514/getCurrentNeighbors-function.txt)

r1de:
do you see the tx_rate and rx_rate for RF type links in your testing?
something is broken... good metrics that were there are now gone. :cry:

ab7pa:
Yes, I see them.  Look at the [sample api output above](https://github.com/aredn/aredn/pull/127#issuecomment-864482224).

r1de:
closing @ab7pa will re-make.

PR #128, opened 2021-06-19
api: revise localhost function

ab7pa:
Revise getLocalHosts() so it doesn't report unwanted hosts.  
getLocalCnxType() is no longer needed.

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -525,16 +525,15 @@ function model.getLocalHosts()
     data = v:splitWhiteSpace()
     local ip = data[1]
     local hostname = data[2]
-    if ip~=nil and hostname~=nil then
-      if hostname:lower() == string.lower( model.getNodeName() ) then
-        cnxtype = "RF"
-      else
-        cnxtype = "LAN"
-      end
+    if ip and hostname then
       local entry = {}
       entry['ip'] = ip
       entry['hostname'] = hostname
-      entry['cnxtype'] = cnxtype
+      if hostname:lower() == string.lower( model.getNodeName() ) then
+        entry['cnxtype'] = "RF"
+      else
+        entry['cnxtype'] = "LAN"
+      end
       table.insert(localhosts, entry)
     end
   end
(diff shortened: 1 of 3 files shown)


r1de:
:+1:

ab7pa:
Everything look the way you were hoping?

r1de:
building and testing it...
looks good reviewing it though. :smile:

ab7pa:
Switched to an alternative to mac2host() to avoid any potential issues with the filenames in /tmp/snrlog.
Now using /proc/net/arp to get the RF mac addresses.  Please test and verify it works as intended.  THX

ae6xe:
Check the HAP aclite and one of the GL models with dual  band radios.  As I recall, wlan0 is the 5GHz AC and only used for wifi-client or lan-ap right now, where as wlan1 is the mesh RF.    Be sure to test on these dual band devices with both radios in use.

ab7pa:
Good point, @ae6xe Joe.  Just pushed a way to ensure the arp query uses the actual RF interface.
Still needs to be tested thoroughly on people's _hAP_ or _AR750_ though.

ab7pa:
Tested the code on dualband radios with active clients on both the Mesh RF and Wifi AP interfaces at the same time.  The **api** returned the correct values for Current Neighbors.  During testing we discovered that _iwinfo_ may not return values for tx_rate, rx_rate, & expected_throughput until there is actual traffic across the link, so we added some error checking in order to avoid a _Bad Gateway_ message in the **api** if that condition is encountered.

r1de:
That sounds like a good solution.

ab7pa:
Discovered that there is a short time while a new _Current Neighbor_ is boot up that nslookup is unable to return proper values.  Added some error checking to avoid _Bad Gateway_ messages from the **api** until those values are available.
   Did additional testing on dualband nodes and verified that the Mesh RF interface is correctly returned and the **api** contains all the correct values for Current Neighbors.

r1de:
:+1: 
Thanks for uncommenting those!
I can't wait for this to get out in the wild and run some new map polling scripts against it.
It'll be neat to see the link lines based on different things besides just the ETX. :smile:

dman776 (approved):

PR #129, opened 2021-06-27
aredn: 20MHz wide channel fix for adjust_rate function in utils.lua

r1de:


r1de:
Looking at this again, dividing `r` by 1 _is_ kinda stupid, _but_ it does make it more "human readable" in a sense.

dman776 (approved):

PR #137, opened 2021-07-20
aredn: fix long node names for tunnels.

### LINKED ISSUE #11: feature: tunnel client character limit warning
Ran into an issue with long node names and tunnels.
It's driven me insane for a week...
Same issue as this:
https://www.arednmesh.org/comment/7412#comment-7412

I did not verify the known limit though. 23 seems reasonable from what I have seen.

There probably isn't an easy way to change this limitation, as it comes from pretty far upstream.
So... I propose, at the very least, a warning on the tunnel pages about this.
At the most, a way to change the "username" that the vtund client uses so that they can be shortened to fit within the limit without having to change the nodes actual name.

r1de:
fixes #11
also fixes: https://www.arednmesh.org/comment/7412#comment-7412
(which has been hanging around for um... _years_!!) :facepalm:

### DIFF
--- a/files/www/cgi-bin/vpn
+++ b/files/www/cgi-bin/vpn
@@ -78,7 +78,8 @@ sub save_clients()
 
         # generate the VTUN NODE name based on the node name and netip
         $net=~ s/\./\-/g;
-        my $vtun_node_name=$parms{"client${i}_name"} . "-" . $net;
+	#VTUN NODE name must not be more than 23 chars long to avoid username limits!
+        my $vtun_node_name=substr($parms{"client${i}_name"},0,23) . "-" . $net;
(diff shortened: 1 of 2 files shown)


r1de:
End users will not notice the truncating that is going on, just use long node names as normal in the tunnel server configuration.
All fixes and work-arounds for this issue should now be null and void. :clap: :clap: 

I tested this on a few "live" tunnels that we use for shaping traffic across the Ventura County backbone links and it worked just fine.
I can see the long, descriptive, node name in the tunnel server config page, but under the hood, `vtun` is using the shortened names for it's "usernames" to login to the other tunnel servers.

This _should_ fix it once and for all!

dman776 (approved):

PR #139, opened 2021-07-21
Api add tactical name

dman776:
add the node's tactical name to the sysinfo block


ab7pa (approved):
Looks good.  Tactical displayed in the **api** just fine.

r1de (approved):
looks good...
I always forget about those hsmmmesh uci settings!

PR #145, opened 2021-09-08
Use develop branch for packages

aanon4:
Switch to using develop branch of aredn_packages so we can include new fixes.


ae6xe (approved):
Not yet any PRs accepted into Arden-packages develop branch, no effect on build for now.

Issue #155, opened 2021-10-31
Disallow web server to use 100% cpu


caseydiers:
During todays mesh-oween, I discovered an issue on my lower memory (32MB) devices.  When there where 1000 nodes all connected, and the device was taxed heavily, they could cope just fine passing traffic at idle.  However, when the webpage was in use it was just too much.  Loading the main page spiked CPU to 100% and loading the mesh status page would timeout several minutes into trying.  It would occasionally drop its connection to the network altogether.

I wonder if the web server software could be limited in the percentage of system resources it can use so that when the device is heavily taxed (as it would be in an emergency) simply loading a webpage doesn't cause the rf link to fail due to the node becoming non-responsive.  Or alternately have the web server system process priority set considerably lower than the other more critical priorities.  I would prefer to have it pass traffic more reliably than load the webpage.  Obviously the solution to all of this is to get a more powerful device, however in this (probably common in these tests) edge case, I believe the software could fail more gracefully.  Especially given that when the connection is dropped, the webpage doesn't load either, so there is no benefit to prioritizing the page load.

Or even a radio box (I don't know how to do a radio box in markdown...:)

- [x] Prioritize Mesh Traffic
- [ ] Prioritize Web Server

dman776:
Which page where you trying to load?

Thanks,
Darryl

caseydiers:
The issue where the webserver uses a lot of cpu is observed in these conditions when loading any of the pages, however most pages load relatively quickly still so it isn't a problem.  However, the mesh status page can take upwards of a minute, so it by far is the largest issue.

What I'm proposing is the 'big hammer' approach of just globally reducing the priority of the web server.  However, I suppose it would also be possible to alleviate the stress by providing an option to not fully load the mesh status page, or somehow reduce the amount of detail, or something, when the memory on the device is full.

PR #157, opened 2021-11-01
Meshoween: Mesh status page optimizations

aanon4:
Based on feedback from the Meshoween event, this diff does a few things:

1. Remove a nested loop which was filtering out neighbor nodes in the "Remote Nodes" section. This scaled O(N^2) as the number of mesh nodes increased and was probably the big reason mesh status ground to a halt on resource constrained nodes.
2. Reduce the size of various in-memory hash tables built when generating the page and discarding them sooner.
3. Detect low memory nodes and limit the size of the mesh status page generated to only closer nodes.
4. For the Remote Nodes section, print as we go. This can get big so let's not keep it in memory.
5. When reading link speeds, reduce reads of the arp tables and wifi files as these are slow.
6. A bunch of little things for small speedups and memory savings.

Note: I haven't written Perl code in 15 years (and then not much) so feedback appreciated.

### DIFF
--- a/files/etc/config.mesh/aredn
+++ b/files/etc/config.mesh/aredn
@@ -11,3 +11,5 @@ config map
         option leafletjs 'http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.js'
         option leafletcss 'http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.css'
         option maptiles 'http://stamen-tiles-{s}.a.ssl.fastly.net/terrain/{z}/{x}/{y}.jpg'
+
+config meshstatus
--- a/files/www/cgi-bin/advancedconfig
+++ b/files/www/cgi-bin/advancedconfig
@@ -163,6 +163,18 @@ push @setting, {
   postcallback => "adjustTunnelInterfaceCount()"
 };
 push @setting, {
+  key => "aredn.\@meshstatus[0].lowmem",
+  type => "string",
+  desc => "Specifies the low memory threshold (in KB) when we will truncate the mesh status page",
+  default => "10000"
+};
+push @setting, {
+  key => "aredn.\@meshstatus[0].lowroutes",
+  type => "string",
+  desc => "When low memory is detected, limit the number of routes shown on the mesh status page",
+  default => "1000"
+};
+push @setting, {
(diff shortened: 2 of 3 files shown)


aanon4:
https://github.com/aredn/aredn/issues/155

ae6xe (commented):
@aanon4 suggest to use, assuming the call exists in perl, get_available_mem() .   I've seen these embedded nodes utilize most of the free memory for performance reasons, and wouldn't expect OOM to be triggered, because there's memory that can be freed up if needed.  I suspect get_free_mem() would return too low a value most of the time.

aanon4:
@ae6xe Will do. The information it scraped from the 'free' command so I'll add a version which uses the 'available' field.

dman776:
Wondering if we should add the thresholds to AdvancedConfig.

aanon4:
Easy enough to add if people want that. Their current values are an educated guess.

ae6xe:
End users generally wouldn't have a concept of what would be a good low memory value.   We could look at what memory would need to be in reserve, that might need to be consumed in parallel to mesh status, and make sure it is available.   This would  be the cron jobs that run every minute.   These jobs go in sequence, so figure out the largest one, plus some additional margin, plus how much mesh status consumes.  Thus could be a minimum setting if allowing adv config to make it a larger value.

aanon4:
Beyond changes based on comments, I've probably done as much with this as I can for now. Here's a couple of datapoints for you all based on some quick testing this evening. The new version scales much more gracefully as the network size increases:

To generate the mesh status page:

SF network (~160 nodes):                                       2.1 seconds (old code).      1.3 seconds (new code)

SF + SOCAL merged networks (~570 nodes):      8.0 seconds (old code).      2.7 seconds (new code)

aanon4:
(... well, apart from a few bugs still to be worked out ..)

caseydiers:
Wow.  This team is fast.  Based on @aanon4 stats above, and the logarithmic reduction in page load times as node count scales, it does indeed sound like it would resolve the issue I opened just a few days ago

propose closes #155

aanon4:
I *really* want to see AREDN scale, so no pesky scale restricting bug is safe :-)

PR #158, opened 2021-11-03
Meshoween: Bug fixes

aanon4:
Collecting bug fixes for the meshoween changes:
1. Create the meshstatus uci section on first-use as it won't exist.


ae6xe (approved):
looks good

dman776 (commented):
this should have been done in /etc/uci-defaults so that it only runs once instead of everytime this page loads.

ae6xe:
Good catch, wasn't thinking about that.

Issue #21, opened 2020-12-16
Hostnames not propagating (was: Nightly Build 960)


ai6bx:
Nightly build 960 seems to possibly have some OLSR issues as nodes names are not consistently resolving to IP addresses and some nodes intermittently drop out as neighbors. I have seen this on Rocket Xw Nanoststion, and PBE 400.

aanon4:
Leaving this here to help inform the next person who sees this:

Node will re-synchronize if sequence numbers change dramatically (ie. on a reboot) and it takes them receiving 16 "bad" messages before it does this (these messages are not propagated so there's no storm impact). Very roughly, a node receives a message from another node about once per-second, so a resync should take about 16 seconds. This is happening all the time when nodes are rebooted. Nameserver message are only sent every couple of minutes, but unless your node is in a particular lossy environment, you'd still expect a name to propagate in under 5 minutes max (let's assume the first name update is lost during resync).

Name information received by nodes persists for 30 minutes, so it seems unlikely for this issue to ever be seen, unless it's two nodes being rebooted around the same time (this information does not persist across reboots).

These pieces of information which would be useful to know when this happens:
1. Does the name eventually resolve? After 5 minutes? After 10?
2. After 30 minutes do other nodes start to forget about the node?
3. Does the IP address of the node still work in the network even if the name does not?

ae6xe:
1) hostname does not resolve, can stay for days, needs a reboot
2) remote nodes do not know histname
3) node is only reachable by IP address by a direct neighbor (need to confirm)

ae6xe:
https://github.com/aredn/aredn/issues/25

Duplicate.

Issue #25, opened 2019-06-18
Hostnames not propagating (was: Nightly Build 960)


ai6bx:
Nightly build 960 seems to possibly have some OLSR issues as nodes names are not consistently resolving to IP addresses and some nodes intermittently drop out as neighbors. I have seen this on Rocket Xw Nanoststion, and PBE 400.

aanon4:
Here's a working theory:

(First I'm assuming OLSR is running in this failure case because you can still reach the nodes via IP address; if it had failed, the routes would expire and the node would be unreachable. Also, the dtdlink come across just fine, which indicates the name server part is also running).

There is code in the name server part of OLSR which strips out invalid entries when it starts up. Invalid can mean a few things, but one of them is an entry referring to an IP address which doesn't exist on a network interface at that moment. Ignoring mid entries (which are only created if the node name is first seen), we have dtdlink entries with one IP address, and the node name with another (the wifi address). So if the wifi IP was missing at a critical startup point, the entries would never be sent.

This explains the scenario described by @k6ccc. The question then is how can this be triggered? This culling process in the name server part of OLSR happens in one of two ways - either the interval time expires or a name message is received from another node. As the interval timer is 30 seconds, that seems an unlikely source. Receiving a name server message could happen at any time after OLSR starts, assuming at least one interface is active, so that is a possibility. Also, bigger networks mean more name server messages, so more chance this can occur at a bad moment.

The big untested part here is, can OLSR be running when the wifi has no IP address yet? There doesn't seem to be any explicit code requiring it to wait for wifi before starting, but there is code to restart OLSRD when network status changes, and perhaps WiFi getting an IP address is one such change? Without understanding more about the NETIFD process (which manages these things) I cant say yet. Perhaps someone knows more about this bit, or has a good way to reproduce this problem?

ae6xe:
As I recall, only dtdlink and tunnel up/down restarts olsr in hotplug, also service advertisement changes.  For timing I did make a change to delay, with no effect:
https://github.com/aredn/aredn_packages/commit/4fe5e8eec06f7900b71a862c645ea374467c7fcd

A manual restart of olsr with a mesh RF link can still exhibit the symptoms.

another duplicate and more info:
https://github.com/aredn/aredn/issues/48

if a node is not in the olsr routing table it will go to default route path, which might reach the node.

aanon4:
[...]

PR #162, opened 2021-11-06
Make sure wifi is up before letting OLSRD run

aanon4:
This is a fix for a long running problem where, very occasionally, nodes don't publish their name until OLSRD is restarted.

The situation happens when OLSRD starts before the Wifi interface is completely up. At startup the name service plugin in OLSRD checks the list of names and services it has configured and removes any which don't have a corresponding, valid, interface. Unfortunately, if an interface becomes valid later, these entries are not added back in. They're gone until OLSRD restarts.

This fix waits until the wifi interface is available and up before allowing OLSRD to startup.

The code is not ideally placed, but I think better here than changing the OLSRD name service to fix the issue, or modifying core packages. I'm open to other options for where this could go.

Should fix: https://github.com/aredn/aredn/issues/21 and https://github.com/aredn/aredn/issues/25

### DIFF
--- a/files/usr/local/bin/olsrd-config
+++ b/files/usr/local/bin/olsrd-config
@@ -56,6 +56,17 @@ if ( $UCI_CONF_FILE eq "olsrd6" ) {
   exit 0;
 }
 
+# Make sure wifi exists and is up to avoid OLSRD nameservice problems later.
+# This is not the perfect place to do this, but it avoids changing more core functions.
+if (system "ubus -t 10 wait_for network.interface.wifi" == 0)
+{
+  for ($i = 0; $i < 10; $i++)
+  {
+    chomp($wifiup = `ubus -S call network.interface.wifi status | jsonfilter -e '@.up'`);
+    last if ($wifiup eq "true");
+    sleep 2;
+  }
+}
 
 @names = @hosts = @services = @tunnels = ();
 


ae6xe (commented):
The wifi interface can be ether wlan or eth.???? (when Mesh RF is not enabled).  This should not be an issue, that I can think of.    If indeed the interface is delayed coming up and olsrd is running -- the conf file exists, then could this code catch a hotplug timing situation?   The hotplug trigger on interface up status will do an olsrd restart -- would this code always be killed on the hotplug restart?

aanon4:
It would be good to detect Mesh RF disabled (although this code will simply timeout after 10 seconds if wifi never appears for some reason).

I *think* the OLSRD config file is regenerated before every restart, so this code will run each time.

What is the hot plug situation specifically? This code is only designed to deal with the edge case where the interface exists but is not fully up before the name server plugin for OLSR deletes entries if thinks are invalid (but become valid moments later).

ae6xe:
@aanon4, I'm not as familiar with the conditions this the wifi interface is wlan or eth physical devices depending if "RF mesh" is enabled or not.

ae6xe:
/etc/hotplug.d/iface/50-olsrd ought to be triggered only when the interface becomes fully up?   If olsr is already up and running with the oslrd.conf file created, then it it looks like this script would restart olsr and should already be fixing the problem.   If I read this hotplug script right, if the interface that just came up, is also in the olsrd.conf file (which means olsr is running), then the script triggers olsr to restart.

Yes, the olsrd config file is recreated at each restart -- ran into this in the past, when I wanted olsr debug options.

ae6xe:
When "Mesh RF" is not enabled, the interface is "eth.3975".  When "Mesh RF" is enabled, the interface is "wlan0" or "wlan1", ...  Always an interface for the main IP address of the node.

aanon4:
As far as I can tell, OLSR and the hot plug system have slightly different ideas of when an interface is fully up. OLSR uses an ioctl SIOCETHTOOL + ETHTOOL_GLINK to determine if the network is up or not, and this appears to give slightly different information to how the hot plug system does it. For wifi, this ioctl only considers the wifi to be up if it has a carrier, while hot plug thinks it's up before that. This is almost always the same, but not quite because the carrier takes a moment to become valid.

I'm making two assumptions though. The first is that the ubus reporting of state is like how OLSR does it, and not like hot plug. The second is that my explanation above is actually what's going on (but I haven't been able to find any other scenario which explains the behavior).

ae6xe:
I'm assuming that olsr does not sufficiently deal with interfaces going down/up, or if the interface is coming up at the same time olsr is starting up.     If olsr adequately handled interface down/up situations, the hotplug 'ifup' shouldn't be needed. 

The changes in this PR are working to handle timing issues when olsr and interface are coming up at the same time.

Might we have the situation where olsr is starting or just barely running the first time, and hotplug is immediately trying to restart it as the interface comes up the first time?

If hotplug code checks to only restart olsr  after the interfaces are checked and confirmed to be up, at least past the code of this PR, then we know olsr is not restarted unnecessarily.   One option,  to touch a file in olsr startup code that hotplug code checks for?  Just checking for the olsr.conf file to exist, is at a time before olsr process is started up on bootup.  

ifup)
                # only work after the first normal startup
                # also: no need to test, if enabled
                OLSRD=olsrd
                [ -e "/var/etc/$OLSRD.conf" ] && {
                        # INTERFACE = e.g. 'wlanadhocRADIO1' or 'cfg144d8f'
                        # DEVICE    = e.g. 'wlan1-1'
                        olsrd_interface_needs_adding && {
                                . /etc/rc.common /etc/init.d/$OLSRD restart
                        }
                }

aanon4:
> I'm assuming that olsr does not sufficiently deal with interfaces going down/up, or if the interface is coming up at the same time olsr is starting up. If olsr adequately handled interface down/up situations, the hotplug 'ifup' shouldn't be needed.

So, OLSRD does handle interfaces going up and down. *However*, the name service plugin for OLSR does not. I suppose the difference is academic really.

> The changes in this PR are working to handle timing issues when olsr and interface are coming up at the same time.

Yes. And in reality, only at boot up as we never drop the wifi link after that without rebooting.

> Might we have the situation where olsr is starting or just barely running the first time, and hotplug is immediately trying to restart it as the interface comes up the first time?

During a normal boot up, OLSRD is started two or three times (restarted once or twice) as configurations are changed (I may try to work out if that can be improved, but another day). So this happens already, but isn't an issue particularly (other than wasting cpu cycles).

> If hotplug code checks to only restart olsr after the interfaces are checked and confirmed to be up, at least past the code of this PR, then we know olsr is not restarted unnecessarily. One option, to touch a file in olsr startup code that hotplug code checks for? Just checking for the olsr.conf file to exist, is at a time before olsr process is started up on bootup.

Sorry - I'm not sure what you mean here :-( Restarting OLSR unnecessarily isn't really a problem, is it?

ae6xe:
Fair enough, re-starting olsr unnecessarily, is a performance  issue, with delay to the end user of when the mesh routing and neighbors become known, maybe up to a 30 sec delay or more depending on interface up triggered restarts.   However, if there is something straight forward we can do to improve the performance by preventing unnecessary delays, this would be beneficial.   What if we simply put in a check so that the hotplug trigger can not happen in the first 1 or 2 min from boot?   The olsr config file would not be changed this quickly (service advertisements, etc.) and hotplug triggers should not be necessary in this timeframe.

aanon4:
> re-starting olsr unnecessarily, is a performance issue, with delay to the end user of when the mesh routing and neighbors become known, maybe up to a 30 sec delay or more depending on interface up triggered restarts. 

Fair enough ... although this change doesn't effect how often or how long it takes to restart OLSRD .. well .. it hopefully pauses it momentarily on first-boot if the interface isn't all the way up ... which is kind of the point.

Possibly there's a misunderstanding here? This change is only querying the state and doesn't make anything happen that wasn't happening already. It's not making any extra hotplug events happen and isn't initiating any OLSRD restarts itself.

ae6xe:
I'm suggesting to add additional code-scope to the PR to address hotplug oslr code performance issue (delay in a node updating to the mesh network routing and services).    I'm happy to commit this as is.  Let me know if you are ready to commit to the master repo, or have any interest to increase the scope of this PR.

aanon4:
Ah - understand (finally!)

Let's not increase the scope of this. I'll take a look at the rest as a separate PR.

k6ccc:
Feedback. Updated a bunch of my nodes to Nightly 642.   I have a few nodes that absolutely every time need an OLSR restart after a reboot - two 5G Rocket XW, and two different Mikrotik LHG.  One of the rockets and one LHG still required the OLSR restart.  The Rockets are DtD connected to the next upstream node and the two LHGs are RF connected.  So far, I have not tried a reboot after the OLSR restart to confirm if the problem returns.  I did do a manual reboot after the firmware update to see if that affected the issue.  No change - still just IP address.

ae6xe:
@k6ccc "One of the rockets and one LHG still required the OLSR restart".  To avoid any doubt, this was after the nightly 642 was installed, correct?     Is there any different configuration or performance of the devices, still exhibiting the symptoms, that stands out?

k6ccc:
Yes, this was after installing Nightly 642.  Nothing unusual.  The two M5 Rockets XW nodes are just sitting in my garage with panel antennas pointed at the sky (no RF connections).  The LHG that needed the OLSR restart is a LHG-2nD-XL that is RF connected to my hAP (about 6 feet away), and it is normally not doing anything other than burning watts.  The LHG that did not require an OLSR restart is a LHG-5HPnD-XL that is RF connected to a Rocket M5 on the roof.  None of those have any extra packages installed.  
I will try a manual restart later this morning to see if the IP only situation returns (had to do some work)...

ae6xe:
The symptoms may be tied to nodes with no RF mesh neighbors.   I have noticed in the past and had people disable "mesh RF" if there are no neighbors to avoid the symptoms -- seemed to be a correlation.   From the earlier dialog, "ioctl only considers the wifi to be up if it has a carrier,".    Maybe olsr continues to think this wlan interfaces is down and doesn't communicate necessary information, and hotplug thinks the interface is up and never triggers the olsr restart, when eventually there is a carrier on wlan interface.   But for there to be a problem, the neighbors would have to be via a different interface dtdlink, if no RF wlan neigbors, so a mystery why the information is not getting communicated fully out dtdlink.

aanon4:
@k6ccc Thank you for trying this ... I just wish it had been successful :-(  As you have what appears to be a very repeatable environment, I'm trying to think of an easy way to test something on your nodes without it being a massive PITA for you. If I make a custom build to try late this week, would you be up for giving it a go?

aanon4:
Also worth commenting - I have an LHG-2nD-XL which is pointing at nothing and was the node which would occasionally fail to restart properly. I hadn't made that connection with this specific model until now.

Orv:
The second Rocket M5 I updated showed the usual symptoms:

642_nightly_build-a

k6ccc:
Yes, I have a very repeatable environment.  The two 5G Rocket XW nodes are future for a local radio site so for now they are just pointed straight up with nothing to talk to.  So all they are doing is sitting there making sure they stay working and getting updates.  The 2GHz LHG is also just sitting there in case I have a field deploy need for it so other than talking to my hAP (6 feet away) it's not doing anything.  More than happy to try test code.
Note that the six Rocket XM nodes normally to not show the OLSR problem.

Orv:
The first Rocket I upgraded did not show the symptoms and it was also a Rocket XM, like Jim's.

ae6xe:
The XW's are newer chipsets, thus higher performance devices than XM.   Is there a correlation, we fully or mostly see the symptoms on the faster devices?

k6ccc:
All three of my Mikrotik nodes (one currently back on RouterOS) show the OLSR problem after reboot most of the time.  The hAP lites do not normally.
And yes, I can fairly easily put the third node back on AREDN if needed for a test.

aanon4:
@k6ccc What the best way of sending you a build to try? I was going too send you one for the 5G Rocket XW and the  LHG-2nD-XL (these are the two types which habitually don't send their names, yes?)

k6ccc:
Correct on the nodes that are habitual offenders.
Either E-mail to jim@k6ccc.org or post to something like Google Docs or dropbox and send me the link.

k6ccc:
Files received.  Updating 4 nodes...

k6ccc:
Looks like success - granted a small sample size.  Both Rocket XWs and both LHGs updated to KN6PLV-nameservice-experiment-a738451 and neither required an OLSR restart.  Checked mesh status page on several different types of nodes that varies from 1 to 6 hops away.  I also was able to reach each node by clicking the node name in another node's mesh status page.  Lastly, I commanded a reboot on all four nodes, and confirmed that all four disappeared from another nodes mesh status page.  Waited a couple minutes, and all four names re-appeared at reasonable ETX numbers...

aanon4:
Excellent. Bit of a good news/bad news situation for me mind you. The good news is that this fixes the issue (in all likelihood as you had it in a fairly reproducible way) and the change I send you confirms my hypothesis about OLSR filtering names for interfaces which aren't "quite" up. The bad news is that I now have to do this "properly" rather than the easy but broken version I tried in this PR, and the sledgehammer test I sent you which simply disables this filter always - harmless but not the intent of the original code.

k6ccc:
One step at a time....
Sledgehammer approach sometimes is needed...

ae6xe:
Shouldn't nameservice be sending the information more than once?   OLSR messages are broadcasts and routinely lost as 802.11n/ac does not retransmit and guarantee a delivery of a broadcast.   It's like nameservice continually thinks the interface is not up or coming up -- more than just a timing issue, rather a stuck issue.

aanon4:
What happens is that the name service plugin to OLSR only checks to see if names and addresses are valid at startup, and discards them from its internal state if not. So it will never send them later regardless of how things change because it no longer knows them.

The real fix is to change OLSR's nameservice plugin to keep all the entries and only send them when they're valid.

ae6xe:
This seems like the issue of hotplug and olsr not detecting the interface state in the same way is contributing to why we see symptoms.  If olsr doesn't see an up or valid interface, then as soon as hotplug detected the interface has come up, it would restart olsr to remedy, but it's not doing that.

aanon4:
> This seems like the issue of hotplug and olsr not detecting the interface state in the same way is contributing to why we see symptoms. 
Yes
>  If olsr doesn't see an up or valid interface, then as soon as hotplug detected the interface has come up, it would restart olsr to remedy, but it's not doing that.
No. In fact its rather the opposite. Hotplug sees the network go up so restart OLSR. OLSR at first doesn't see the network as up (and so strips the "bad" options from its internal name service state). A moment or two later OLSR does see the network go up, and so goes about its usual business as one would expect (sans those now removed name services entries). Bother hotplug and OLSR behaved correctly. Their view of the network state isn't identical but overlaps. Most of OLSR is completely fine with this discrepancy, but the nameservice wasn't written to handle it. The build I sent to @k6ccc removes the filtering stage from the name service, and the problem appeared fixed. I've just sent him a "better" build (ie. one that fixes is correctly rather than crudely as in my test build) so hopefully we'll see how that holds up.

k6ccc:
Just got home and saw the new versions.  I will load them shortly...

k6ccc:
All four nodes updated and working properly.

aanon4:
Excellent. I've published a new PR based on these changes: https://github.com/aredn/aredn_packages/pull/6

PR #164, opened 2021-11-06
update status page to add additional location information 

wh6av:


dman776:
For consistency, I prefer that we keep the logo in the upper left

dman776 (changes requested):
Setting should be stored in uci format moving forward.   Put devicelocation in the etc/aredn file and make sure that you add to uci-defaults as well.

dman776:
Also, the devicelocation should also be added to the /www/cgi-bin/api (and don't forget to increment the api version number.)

r1de:
Don't get me wrong, I'm all for more info to be shown, but isn't the lat/lon already shown on the main page (if set)?
And we already have the node description field, would that not be used for a more generic location? ie: `This device is located in Oxnard, Ca.`

Not against the change, just sayin' that it seems redundant is all. :smile:

dman776:
Fair point

dman776:
Any other community comments on this??

wh6av:
Sorry, been busy with work here.  I'll get to it this week.
image

dman776 (changes requested):
needs work to comply with uci storage of info.   remove css changes

dman776:
closing.   no response.  Also, changes would need to be based on the new LUA version of this page.

PR #170, opened 2021-11-10
fix: duplicate to the config.mesh structure after saving.

dman776:
fix issue of lat/lon clearing after a Save Settings.

### DIFF
--- a/files/etc/config.mesh/aredn
+++ b/files/etc/config.mesh/aredn
@@ -13,3 +13,6 @@ config map
         option maptiles 'http://stamen-tiles-{s}.a.ssl.fastly.net/terrain/{z}/{x}/{y}.jpg'
 
 config meshstatus
+
+config location
+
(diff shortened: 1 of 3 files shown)


ae6xe (approved):
As discussed.  Looks good.

k6ccc:
Re:  fix issue of lat/lon clearing after a Save Settings.

This fixed it all right.  Out of 13 nodes updated from 648 to 654, every single one lost lat / long information.  I had to manually re-enter it after the update to 654 and was able to save it on the first attempt.
This included Rocket M3 XM, Rocket M5 XM, Rocket M5 XW, GLiNet AR750 and USB150, Mikrotik hAP and both 2GHz and 5GHz LHGs.

k6ccc:
Update to my last.  After entering and saving the Lat / Long, it does show on the main status page, but not on the Mesh Status page.

PR #171, opened 2021-11-10
bugfix: location migration

dman776:
explicitly set the config.mesh directory for changes with uci command, otherwise, no changes are made.
to test:
- downgrade to nightly build prior to 654
- set location and gridsquare using the UI
- remove 'config location' sections from /etc/config/aredn and /etc/config.mesh/aredn
- upgrade

Values will be migrated to the above files and /etc/latlon and /etc/gridsquare will be removed


aanon4:
I put the current release build on my test node, make sure the location was correctly setup. Upgraded with this PR build ... and the location vanished. No latlon file anymore and no location information in either /etc/config/aredn or /etc/config.mesh/aredn.

aanon4:
More: I reset the lat/lon and they are correctly stored in the aredn files in both /etc/config and /etc/config.mesh. I then rebooted and they're gone again. It looks like the **96_aredn_migrate_latlon** file doesn't terminate correctly, possibly because I didn't have a /etc/gridsquare file because this wasn't set on the node.

files/etc/uci-defaults/96_aredn_migrate_latlon
```diff
@@ -0,0 +1,28 @@
...
+/sbin/uci -c /etc/config.mesh -q set aredn.@location[0].lat="$LAT"
+/sbin/uci -c /etc/config.mesh -q set aredn.@location[0].lon="$LON"
+/sbin/uci -c /etc/config.mesh -q set aredn.@location[0].gridsquare="$GRIDSQUARE"
+/sbin/uci -c /etc/config.mesh -q commit aredn
+
+cp /etc/config.mesh/aredn /etc/config/aredn
+
+# remove the /etc/latlon and gridsquare files
+rm /etc/latlon
+rm /etc/gridsquare
```
aanon4, line 28: Add '-f' flags here because these files might not exists. Without -f this script with terminate with an error (causing it to be rerun on the next boot .. and the next .. etc.)

dman776:
Thanks Tim.  Will update late tonight.

dman776:
started with older nightly with lat/lon/gs.
cleared gridsquare and saved.  works as expected.
rebooted node.  still has lat/lon as expected.
status page displays lat/lon correctly.

aanon4:
Still issues I'm afraid :-( This was my test scenario:
1. Started with a node with the current release build. Lat+Lon set, no grid square.
2. Upgraded to a build based on this PR. Succesfully migrated Lat+Lon to the UCI System. The old /etc/latlon file was removed.
3. Rebooted. Node still fine.
4. Upgraded the node again (a test build of my 'stop_multiple_alerts' PR rebased from this PR, so this is included as if it was merged into main). Locations are now missing and are absent from the UCI System.

What I believe is happening is that the 96_ uci-defaults file is being run during the second upgrade (all uci-default files are run after every upgrade). It won't find the /etc/latlon file this time around, so LAT/LON aren't set. But the UCI set commands are still executed, deleting the values from the UCI System (setting them to nothing).

PR #172, opened 2021-11-11
Stop creating 'config alerts' on every reboot.

aanon4:
A 'config alerts' section was being added to the config on every reboot because this check will always be true if an alerts URI hasn't been set. Now it will only get added if it doesn't exist which seems more correct.


dman776 (approved):
tested ok on hapaclite nightly.   no additional config alerts sections where created

PR #174, opened 2021-11-11
bugfix: api traceroute function hardening

battlehax:
Prevent unauthenticated remote code execution as root in the 'traceroute' function. This could probably be improved, or the function removed entirely, or to apiprotected. This affects all nightly/custom builds since June when this function was added.


dman776 (commented):
good catch.

dman776 (approved):
tested and approved

PR #175, opened 2021-11-11
Prevent mesh from accessing internal LAN

battlehax:
With WAN sharing enabled, users across the mesh could access your internal LAN devices if your mesh device is plugged into another router or network to get internet.


ae6xe (changes requested):
@battlehax,   This is historic long before anyone's time still making changes to the firmware, however the files in /etc/config/ are built after doing a "save settings".   These  PR changes need to go to /etc/config.mesh/firewall location.  This config.mesh/ location is a template where files are copied over to /etc/config/ and there's notation in some config files to insert settings.  For example, the IP address is calculated randomly based on a MAC address for each node, thus in the network config file, we see notation where this information will be inserted.     The /etc/config/firewall config settings are only used on firstboot, before the user has configured for AREDN and saved the settings, overwriting /etc/config/firewall.

Also, there is a way in the UI on the "Port Forwarding, DHCP, and Services" page to create firewall port forwards from a node's LAN device to a WAN IP address, e.g. allow access to a home printer.   I don't recall the order of the firewall rules, but we need to confirm, if anyone has one of these firewall settings created in the UI, that these occur in the firewall chain-logic first, before everything is dropped.   I don't recall but a user adding a customer rule, may only get appended to the firewall config file at the end, and the drop rules added in this PR would go first, never reaching the custom rules.

battlehax:
Ok, I'll work on this. Personally, I've not gotten the port forwarding to work from the webui. I'll move things around and do some more thorough testing,

ae6xe:
I don't recall if port forwarding UI custom settings have been tested or even used by anyone in a very long time.   Possibly, no news is good news, surely someone has tried using.

Also, take a look at /etc/local/mesh-firewall/  .  This may be another option to implement these settings.   Although, this location is intended to load firewall rules when a package is loaded and needs new rules.

ae6xe (changes requested):
@battlehax , it's a bit tricky to put the rules in the right place in the openwrt firewall structure.  Right now the forward chain looks like this:

Chain FORWARD (policy DROP 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination
 490K  173M forwarding_rule  all  --  any    any     anywhere             anywhere             /* !fw3: Custom forwarding rule chain */
 472K  163M ACCEPT     all  --  any    any     anywhere             anywhere             ctstate RELATED,ESTABLISHED /* !fw3 */
    0     0 zone_vpn_forward  all  --  tun+   any     anywhere             anywhere
18174   10M zone_lan_forward  all  --  br-lan any     anywhere             anywhere             /* !fw3 */
    0     0 zone_wan_forward  all  --  wlan0  any     anywhere             anywhere             /* !fw3 */
    0     0 zone_wifi_forward  all  --  eth1.3975 any     anywhere             anywhere             /* !fw3 */
    0     0 zone_dtdlink_forward  all  --  eth1.2 any     anywhere             anywhere             /* !fw3 */
    0     0 reject     all  --  any    any     anywhere             anywhere             /* !fw3 */

There's already a reject at the end of the FORWARD chain that should be triggered if the logic got that far.   Additional rules would be ignored.  on a mesh node do the command "iptables -L -v" and follow through the chain of rules -- there will be a right place to insert the commands, and the -o option would be one of the defined firewall zones (a specific hardware interface could be different on different devices).   The logic stops at the first rule that matches the packet.  it's been a while since I looked at these rules, thus I don't recall offhand the exact spot to put them in.

battlehax:
Sorry, didn't realize my changes made it here. I've been bashing my face against the keyboard all weekend with this and getting nowhere. I won't push anything back to my fork for now.

ae6xe:
we can collaborate offline:  ae6xe at arrl dot net to get this figured out.   It's been a while since I dug through the openwrt firewall rules, but will come back quickly.  Maybe I can help jump start getting familiar and figuring out where to put these rules in.

dman776:
Once this is ironed out.... wouldn't this be better as an optional package that ops could elect to install if they want it?   I can see some situations where it would be desirable to access the "upstream network".    Ie. A node at the EOC may want to allow the mesh to access other systems at the EOC

battlehax:
Honestly, I feel it'd be better as a default that can be overridden by custom "port forwarding" rules. Consider that a malicious party could simply ignore FCC regulations, create a rogue node, and attack nearby neighbor's internal networks.

dman776:
@ae6xe any comments on this PR?

ae6xe:
From my perspective, the risk-impact that a mesh network might be infected with the next RansomWare threat is probable over time.   Not all mesh users are keeping their devices current with OS patches and AntiVirus.  This change will protect and segment all user's home networks from lateral propagation.   It is a reasonable tradeoff to burden users to add point forward rules from the mesh into their home network, e.g. a printer, file share, etc. to protect everyone's home network from future threats.   To accept this PR, we might follow up to validate the port forwarding rules in the UI still work, and consider a way to expand and improve capability and usability.

KM6IAU:
This is difficult because, while a person setting up a gateway ought to
understand routing, and while a node with a local internet gateway ought to
be isolated by firewall from a private LAN tangent to the WAN, we can
expect that some fraction of users will not understand even the basics of a
discussion about routing.

As much as I would like to see stations with a local internet gateway be
isolated from private networks with an appropriately configured
multiple-interface router, it is probably a better solution to firewall the
private IP space and have this feature able to be disabled in Advanced
Configuration.

As an example of why one might want to disable it: they have their own
properly configured router, and also have a pihole DNS server answering
requests for internet-bound mesh traffic on the private subnet.

dman776:
I think having this a something that users can disable via AdvConfig is a good compromise.   It can be enabled as the default, but, for those who dont want this behavior, they can disable.

battlehax:
Sorry I've been away. Still working on the how, and making some progress. Things have been busy here, but I have not given up, as this is (to me and hopefully you) a very important security risk!

On the why, I still stand that being able to access an rfc1918 address from inside the mesh shouldn't ever need to happen. If people need to access assets behind the nat, that's what forwarding rules should be for. If you think you need to have your rfc1918 address accessible inside the mesh, consider this; accessing 192.168.x.x from inside the mesh might not work the way you expect it to, depending on where you are inside the mesh, that could be a completely different person's internal network (currently whoever is closest with wan enabled) so using bookmarks or sharing links to assets will not work the same across different parts of the mesh.

KM6IAU:
That is a good point. I think perhaps there is some confusion of intent.
 I agree that rfc1918 addressed should never need to be accessible from the
mesh--the obvious exception being the gateway itself... but a less obvious
exception would be providing DNS resolution, or NTP services.

I have a couple nodes, KM6IAU-Onaga-Node1 (10.13.81.1) and
KM6IAU-Onaga-Node2 (10.13.82.1).
Both nodes are advertising themselves as internet gateways, using the IPs
*172.16.73.1*/24 and *172.16.73.2*/24, respectively.
Their gateway is *172.16.73.254*, which the eth8.1 (eth8 vlan1) interface
of my router.
Their DNS1 and DNS2 are also set to *172.16.73.254*.
My router then provides "Captive DNS" by DNATing ANY traffic to port 53
(DNS) to 172.16.0.73 (a PiHole, on another interface, providing ad
blocking, and forwarding to Cloudflare's 1.1.1.3 for further malware
protection, and family-safe DNS resolution).

Now, while my router is capable of DNATing DNS and NTP, and thus *my*
internal IPs are irrelevant (from the perspective of internet-bound mesh
traffic), other users may want to use a PiHole DNS on their LAN (which is
tangent to the WAN), and skip fancy DNAT tricks.

So, the node itself needs to be able to access rfc1918 addresses, but I
agree, there is probably no good reason to allow rfc1918 beyond the
particular AREDN node acting in the role of internet gateway.

Forgive my verbosity for clarity's sake.

Issue #156, opened 2021-10-31
Have Firmware Update section of Admin page optionally show nightly firmware instead of stable


caseydiers:
I'm curious if there is interest in this feature.  If so, I will dive in and work on it.

### Feature  Request:

Currently the software will download and display the stable version of the software.  What I'm proposing is that there is an option in "advanced options" to display nightly builds in the Firmware Update page to simplify the update process for those the typically keep their devices on the nightlys update.

The folder structure for the stable releases (http://downloads.arednmesh.org/firmware/) isn't exactly the same as the nightlys (http://downloads.arednmesh.org/snapshots/trunk/targets/) however, I think it would be possible to:

- change the folder to look in from default to nightly
- dive into the correct folder, for example, from http://downloads.arednmesh.org/snapshots/trunk/targets/ to http://downloads.arednmesh.org/snapshots/trunk/targets/ar71xx/generic/ based on the information located in the hardware type field on the admin page
- locate the correct firmware type based on the sha256sums info located in that folder, or alternately by just parsing the files that are in that folder
- do the rest of the process the same as the default release.

Obviously this would be easier if the build system could generate a .list file for the nightlys as exists in the actual release folder, but it doesn't look like that is necessary to make this work.

If the project was in PHP I'd have contributed already, but I'm ready to pull up my sleeves on a couple of improvements in the UI and learn a bit of perl along the way.  From what I've looked at so far, it isn't too much different from what I'm used to so I should be able to pull it off.

caseydiers:
KV3T here btw

KM6IAU:
I'd love this.

I just started developing a new webUI.  I share your sentiments about the Perl.  I'd rather it was Python.  The further I get into developing the new webUI, the closer I get to just rewriting it in Python.

I will follow this feature request, and include it in my work.

dman776:
Check out the aredn_next repo.  That is the basis for our next generation ui

PR #178, opened 2021-11-14
Display firmware update prompt

### LINKED ISSUE #17: Enhancement Request: Display a text on node homepage when newer firmware version is available
This message is an enhancement request for the homepage of each node, to display a message when a new stable and/or nightly firmware newer than the one running is available.

As a suggestion on how this could be done: 

Create a RES API URL returning the latest stable versions, and latest nightly build versions.
For example https://www.arednmesh.org/api/latest_firmware/ , returning something like:
{
    "ubnt-bullet-m": [ { "stable": "3.30.4", "nightly": "1105" } ],
    "ubnt-nano-m":   [ { "stable": "3.30.4", "nightly": "1105" } ],
    "ubnt-rocket-m": [ { "stable": "3.30.7", "nightly": "1107" } ],
    ...etc...
}

This can then be queried by code on the node homepage, compared against current version, and
* if current_version < latest_stable then display "new stable version available !"
* if current version < latest_nightly then display "new nightly version available !"

I am sure this is very simplistic and might require significant changes in the versioning mechanism, just wanted to plant a seed for future revisions.

Will be happy to help in any way I can.

Thanks !

caseydiers:
Here is my first pull request to the project.  It fixes #17 and puts some building blocks in place for #156.  This patch adds some code to the status page to compare the current firmware number to that listed as the current firmware number in the .list file downloaded from the website.  If that file has not been downloaded since the last reboot, it will load the rest of the status page and then download the file.  If the file exists but is older than 7 days, it will download the new version.  If it discovers that a new version is available, it will indicate as much on the status page.  If it discovers that you are on the current firmware, it will add a green check mark next to the firmware number indicating that you are on the newest version.  Since a lot of the users use the nightly firmware versions, I added a check box in the advanced page to indicate that this node is using nightly files.  When this box is checked, it won't tell you to "update" to the older stable files, it will simply indicate that nightly files are in use, and it isn't therefore checking for new firmware versions.


files/www/cgi-bin/status
```diff
@@ -51,6 +51,42 @@ $wifi_iface = get_interface("wifi");
...
+$fw = `cat /etc/mesh-release`;
+$nightly = `/sbin/uci -q get aredn.\@downloads[0].nightly`;
+
+# Check to see if the node is on a nightly build, and if not proceed in loop
+if($nightly != 1) {
+
+	# Check if there is an update to the firmware
+	if(-e '/tmp/web/firmware.list') {
+
+		# See if new file should be downloaded, and if so, set a variable to do this at the end of the script
```
dman776, line 63: i think this logic would be better in a separate script that is called by a cron job instead of everytime the status page is loaded.

dman776 (commented):
FYI, the process will be changing for how the download files are handled.   We are moving away from these .list files into build generated json file(s) which will include nightly builds in it.  I think we could add this supporting function of "get list of firmware updates" into the api so that when the process changes, it is in the underlying layer of the api.   Also, we will be able to leverage the api in the new user interface that is being developed.

files/www/cgi-bin/status
dman776: (sorry, wrong code snipped selected.  was referencing the date checking code.)

caseydiers: My thinking was that the status page, even on active nodes, probably isn't loaded all that often, and that info is really only valuable when the page is being loaded.  A cron job would have it run all the time in the background, whereas this will only check when there is a user that might actually view and/or do something with the info.

As a more concrete example, if a node is put up and no one ever logs into it but it is very busy passing traffic away in the background, the way I wrote this it would have no impact whatsoever on the node nor the aredn website.  However, if a user logs in and starts navigating around, it will check once and only once, if there is an update, and prompt the user accordingly.  If they then don't log in again for another year, again no impact for another year.

It isn't a ton of traffic / bits, but every time it pings home, it is downloading that file from the AREDN servers as well.  This way it would, at most, download the file once a week, and at least it would never download the file at all.  And when the download does happen, it would be at a totally random time.  If we did a cron job, you might see 1000 hits to that file every Monday morning at 4am west coast time for example, even if no user will see that info before the next update on any given node.

If a cron job is preferable by the team that would be a relatively easy change, however the logic I noted above was why I believe the current setup (or something like it) is preferable to a cron job.

dman776: agreed.  good point.

caseydiers:
My contribution appears to be have been untimely.  I believe this contribution is no longer relevant to the project, so I'll close the pull request.  Better luck next time, I suppose.

PR #179, opened 2021-11-16
Perl to Lua rewrite - part 1

aanon4:
[DO NOT MERGE]

Time to make all there Perl go away (and free up the space its taking).
Very much here for feedback.
This will build and run (at least as an update - don't know about a fresh install) correctly.


files/usr/local/bin/mgr/utils.lua
```diff
@@ -0,0 +1,103 @@
+
```
dman776, line 1: for the mgr.utils, i think some of these are duplicated in /usr/lib/lua/aredn/utils.lua.  why not just require("aredn.utils")?

aanon4: mgr.utils is a bit of a dumping ground at the moment (and yes, some duplicates or near duplicates in there too). I don't expect it to survive to the end.

ae6xe (commented):
did the old perl scripts get removed from /usr/local/bin in this PR?

suggest we do a round of testing prior to putting in nightly build .

There have been linux and probably 'iw' upgrades since the iw zombie handling was put in.   To know if this logic is still needed, we'd need to ask the community to supply the /tmp/zombie.log files across their (lengthy uptime) nodes to see if this is still actually killing something.   I'll put a forum post, to ask the community to send in this information.

aanon4:
I've left all the old scripts in place, but did change the setup (e.g. crontab) so that both versions of replaced scripts don't run at the same time.

I think this requires a lot of testing before it hits even nightly - it's here mostly for feedback at this point. I don't recommend anyone should rush out and test it just yet (it should work mind you, it's just probably premature for that kind of operational feedback)

Be nice if the 'iw' problem had gone away.

ae6xe (commented):
init.d manager could be hardened if want to put in this initial commit.     ref: https://openwrt.org/docs/guide-developer/procd-init-scripts  in case you've not already bumped into.

init.d/olsrd script example -- can quickly add way to respawn, etc. 

regarding linkled.lua:   possibly the led link path can be pulled from /etc/board.json on every device now days.   Opportunity to clean up this code.  We shouldn't have to figure this out on every device.

aanon4:
@ae6xe so I see led info in /etc/board.json for leds but I'm not sure what the general solution is for "linkled.lua"? Should I always use the "led.rssilow" entry? (if you don't know I can work it out, but I hoped you did so I didn't have to :-))

ae6xe:
The led lights are used this way:  https://www.arednmesh.org/content/led-lights-0 .     The light unique to aredn has been the "RF low" light -- used to indicate there is an RF Mesh neighbor link or dtdlink neighbor (don't recall if this lights if only a tunnel neighbor).  Otherwise, openwrt/linux default led light setting.  We don't use the led lights to show Signal strength.   As far as I know, we should pull directly from board.json and use led.rssilow.  The reason we started hardcoding this elsewhere is because the device wasn't supported at the time by openwrt, and this detail didn't exist in board.json (where we should have added it for a new device :) ).   it's possible some devices, possibly the hap ac lite, don't have an rssilow led on the device, trying to recall.  if so, nothing to do.     If we do find in the future a device doesn't have rssilow in board.json, and should, then we would fix it there and not continue a hack to work around elsewhere.

dman776:
Joe and I were chatting today.   We should no longer need this "radio.json".  We plan to create an /etc/aredn.json at firstboot that pulls all the needed info from he device.  Still need to determine how to do the "supported" flag if we want to keep that functionality.    ie.
`{
     "wifi": [
          { 
               "name": "phy0",
               "device": "radio0",
               "ifname": "wlan0",
               "band": "2400",
               "channels": [
                    { 
                         "num": -2,
                         "freq": 2397,
                         "power": 22,
                    },
                    { 
                         "num": -1,
                         "freq": 2402,
                         "power": 22,
                    },
                    { 
                         "num": 1,
                         "freq": 2412,
                         "power": 22,
                    },
               ],
          },
          { 
               "name": "phy1",
               "device": "radio1",
               "ifname": "wlan1",
               "band": "5",
               "channels": [
                    { 
                         "num": 36,
                         "freq": 5397,
                         "power": 30,
                    },
                    { 
                         "num": 37,
                         "freq": 5305,
                         "power": 28,
                    },
                    { 
                         "num": 38,
                         "freq": 5300,
                         "power": 24,
                    },
               ],
          },
     ],
}`

aanon4:
> Joe and I were chatting today. We should no longer need this "radio.json". We plan to create an /etc/aredn.json at firstboot that pulls all the needed info from the device. Still need to determine how to do the "supported" flag if we want to keep that functionality (i think this becomes less of an issue with the ath79 arch as each device gets it's own bin file). ie.
> `{ "wifi": [ { "name": "phy0", "device": "radio0", "ifname": "wlan0", "band": "2400", "channels": [ { "num": -2, "freq": 2397, "power": 22, }, { "num": -1, "freq": 2402, "power": 22, }, { "num": 1, "freq": 2412, "power": 22, }, ], }, { "name": "phy1", "device": "radio1", "ifname": "wlan1", "band": "5", "channels": [ { "num": 36, "freq": 5397, "power": 30, }, { "num": 37, "freq": 5305, "power": 28, }, { "num": 38, "freq": 5300, "power": 24, }, ], }, ], }`

Certainly would be better than loading up the whole "database". radio.json is just a copy of the perl version. Easy enough to change once this exists.

dman776 (commented):
Tim, for the files that are being PORTED from Perl to Lua, make sure that you include the prior author's copyright line in the source headers since this is a derivative work.

ie. 
Copyright (C) 2021 Tim Wilkinson
   See Contributors file for additional contributors
  Copyright (C) 2015 Conrad Lara
  Copyright (c) 2013 David Rivenburg et al. BroadBand-HamNet

dman776:
I tested the usr/lib/lua/aredn/info.lua lib and supporttool.lua today.  works as expected.

aanon4:
> Tim, for the files that are being PORTED from Perl to Lua, make sure that you include the prior author's copyright line in the source headers since this is a derivative work.

Done!

aanon4:
Okay. Moving on to real PRs now so closing this.

PR #180, opened 2021-11-16
undo the renaming of output files

dman776:
that previously removed the -squashfs- portion to accommodate the OFS tool

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,15 +150,6 @@ compile: stamp-clean-compiled .stamp-compiled
 	  -name sha256sums -o -name "*.buildinfo" -o -name "*.json" \) \
 	  -print \)`; do rm $$FILE; \
 	done;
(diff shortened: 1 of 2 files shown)


ae6xe (commented):
This change also removes the "rm" command that removed other informative files.    As I recall, these were removed to not cause confusion to end users.   I see this as a minor concern, so taking out this complexity works for me.

With ath79, the image names are 1x1 mapping or nearly 1x1 mapping and we may need to update this admin script shortly to accommodate. 

Any changes downstream in build or on node checking for correct firmware?

dman776:
the removing of "extra" files is not included in the PR and is still in the Makefile just above this section.   As far as other impacts, the only node code that I see that this is the included admin page.   The gen5_list.pl Downloads page generator script that runs on otis is also impacted, but, will no longer be needed once OFS is implemented.

ae6xe (approved):
looks good.   ...misread "mv" as "rm" :) .

PR #189, opened 2021-12-10
Lua Services

aanon4:
Migrate a set of services currently in Perl and SH into Lua, and mange using a central Lua manager to better coordinate the operation and to reduce overall memory footprint.


aanon4:
This is the first of the two PR's we discussed. If someone who isn't me could spin this up on a node to sanity check that'd be fantastic.

ae6xe (commented):
rssi_monitor implementation is only handling the horizontal receive chain.   The original also handled the vertical receive chain.  The use cases include single chain devices that generally will be on the vertical polarity, e.g. a bullet with an omni antenna.    The current implementation would not cover this use case.

aanon4:
I had meant to ask you about that but forgot. The Lua API provided only a single piece of information, which I'm assuming (big assumption) is a combined value and that'd work okay? If not I can pull the information the old way like the perl code did.

ae6xe:
I don't know how well the combined value will work.   I know that the symptoms were seen on only one polarity-chain.   If the values from 2 chains are combined, there would be less sensitivity, so we risk it won't trigger when it needs to.  This appears to be a tradeoff of more attention on testing with the current code vs adding the code to handle both receive chains.

aanon4:
I will just go with a two-chain solution because we know that works - thanks!

dman776:
adding the first of the dump_perl project as discussed

PR #194, opened 2021-12-17
remove 'trunk' from snapshots path

dman776:
standardize the nightly build/snapshots directory paths to remove the 'trunk' from /snapshots/trunk/
in support of the AREDN Firmware Selector tool
Also need to modify the CircleCI environment variable NIGHTLY_DIR before merging


ae6xe (approved):
Confirmed all the places to make the change.

Issue #196, opened 2022-01-02
Lat/Lon not showing on Mesh Status page


n2mh:
Upgraded to Nightly 739 from various earlier versions - 4 nodes showing this issue
The latitude and longitude are not showing on the Mesh Status page but do show on the main Status page. Reviewing /etc/cgi-bin/status and /etc/cgi-bin/mesh shows a difference in how lat/lon is retrieved.

73

n2mh:
Additional info: these nodes are all hAP-Lites.

n2mh:
Also affecting a Bullet Titanium on 2GHz.

k6ccc:
Confirmed on:
Mikrotik hAP AC lite
Mikrotik RB-LHG-2nD-XL
Mikrotik RB-LHG-5HPnD-XL
Ubiquiti Rocket M3 XM
Ubiquiti Rocket M5 XM
Ubiquiti Rocket M5 XW
GL-iNet AR750
GL-iNet USB150
That's every node type I have...

VA2XJM:
It would be interesting to copy/paste your sysinfo.json content so it can be compared and see if the format have changed...

sysinfo.json is accessible via http://YOURNODE:8080/cgi-bin/sysinfo.json

Here is mine on latest release and working perfectly with MeshMap:
`
{
   "lon": "-77.7861",
   "sysinfo": {
     "uptime": "172 days, 19:40:27",
     "loads": [
       0.52000000000000002,
       0.42999999999999999,
       0.40999999999999998
     ]
   },
   "interfaces": [
     {
       "name": "tun51",
       "ip": "172.31.227.14"
     },
     {
       "mac": "B8:69:F4:D7:81:E9",
       "name": "br-lan",
       "ip": "10.208.61.33"
     },
     {
       "name": "wlan0",
       "mac": "B8:69:F4:D6:81:E8"
     },
     {
       "mac": "B8:69:F4:D6:81:E9",
       "name": "wlan1",
       "ip": "10.214.129.233"
     },
     {
       "name": "tun52",
       "ip": "172.31.227.18"
     },
     {
       "name": "tun50",
       "ip": "172.31.227.10"
     },
     {
       "name": "tun53",
       "ip": "172.31.227.26"
     },
     {
       "mac": "B8:69:F4:D7:81:E9",
       "name": "eth1.2",
       "ip": "10.215.129.233"
     },
     {
       "name": "tunl0",
       "mac": "00:00:00:00"
     },
     {
[...]

PR #197, opened 2022-01-03
Fix missing location

aanon4:
Change the way location is retrieved in the mesh page to the same as the status page

https://github.com/aredn/aredn/issues/196


ae6xe (approved):
code review only. looks good.

PR #201, opened 2022-01-16
bugfix: remove filename checks for firmware uploads

dman776:
remove marginally effective filename devicetype checks.


r1de (approved):
tested and works, filename check gone.
went back and forth with original file several times.

ae6xe (approved):
tested on hap ac lite with success.

PR #204, opened 2022-01-19
build: add AFS collector step to pipeline

dman776:
adding a build pipeline step to collect all firmware files for AREDN Firmware Selector tool (AFS)


ae6xe (approved):
I'm not familiar with collect.py, but otherwise it looks good tome.

Issue #206, opened 2022-01-20
NB811 aredn alert message issue


ab7pa:
On today's NB811 the perl webpage takes into account that a zero-length alert message file may exist, but the lua page does not do that.  The result is a blank placeholder for local messages.
<img width="235" alt="zero length local message file" src="https://user-images.githubusercontent.com/69524416/150409304-e5806057-e43a-473a-9fea-1c8df73a214f.png">
<img width="621" alt="empty placeholder" src="https://user-images.githubusercontent.com/69524416/150409366-03761c6e-c98f-4410-ad8e-0ce1e7169cf1.png">

ab7pa:
This has been fixed in subsequent nightly builds.  THX

PR #207, opened 2022-01-21
Fix Issue #206 - Displaying empty alerts

aanon4:
Now ignore empty alerts.

https://github.com/aredn/aredn/issues/206


ae6xe (approved):
looks good.

Issue #205, opened 2022-01-20
NB811 mesh status layout issue


ab7pa:
I noticed on today’s NB811 that the layout of the Mesh Status page is slightly different.  If this was not intentional, then maybe it should be looked at in order to have it correspond with the perl version.
![new mesh status layout](https://user-images.githubusercontent.com/69524416/150409048-66194911-2b27-4907-8e9c-10b18e9b913e.png)

aanon4:
If you add .pl to the url you can get old old perl version of this page. Can you post that for comparison?

ab7pa:
mesh.pl for comparison
<img width="1165" alt="mesh.pl comparison" src="https://user-images.githubusercontent.com/69524416/150445654-944d324b-76e0-49d0-881f-c60fcecfdf91.png">

ab7pa:
This has been fixed in subsequent nightly builds.  THX

PR #208, opened 2022-01-21
Fix issue #205 - Bad local service layout

aanon4:
Fix local service layout (missing piece of markup)

https://github.com/aredn/aredn/issues/205


ae6xe (approved):
looks good.

Issue #211, opened 2022-01-24
Validate Link led for all board.json files for all supported devices


aanon4:
Validate all the board.json files to make sure we have a valid Link LED we can flash.

Issue #203, opened 2022-01-18
NB798 led issue on USB150


ab7pa:
The GL.iNet USB-150 only has two leds.  The bottom one indicates power, and the upper one should indicate when the USB150 is associated or meshed with other nodes.  In my recent testing, the upper led remained unlit even when the unit was associated/meshed.

dman776:
need to check if we have access to that led

K6AH:
My USB150 does exactly the same thing.  Works as it should on 3.22.1.0, not on NB798.

dman776:
Ahhh.   Could be the lua based linkled function

Thanks,
Darryl

aanon4:
I'm sure. We changed how the LED light was determined in the lua code compared to the older perl code, simply pulling a specific led definition from the board.json info. It looks like this isn't going to work, or at least cause too many regressions, so I'm going to revert that code back to the original (which was just a copy of what Perl did).

ab7pa:
NB838 fixed this issue.  USB150 link light now illuminates when linked.  THX

PR #212, opened 2022-01-24
Link led fixes

aanon4:
* Flash the LED slowly when no link found.
* Handle the LED exceptions where there is no 'correct' definition in board.json
   We should fix the board.json creation process to add the correct link led definition, but for now just reflect the
   exceptions currently in the Perl code. In fact, these aren't necessarily correct (the one I checked wasn't) so it
   probably requires someone with access to devices actually making sure we do the right thing.
   Spreadsheet of leds according to the current build code: https://docs.google.com/spreadsheets/d/1-kUm4GpUwlrVnHdr4xzSoVxGtPmMZS8x/edit?usp=sharing&ouid=104861343244109655077&rtpof=true&sd=true
* See https://github.com/aredn/aredn/issues/203
* Also https://github.com/aredn/aredn/issues/211

### DIFF
--- a/files/usr/lib/lua/aredn/hardware.lua
+++ b/files/usr/lib/lua/aredn/hardware.lua
@@ -134,26 +134,59 @@ function hardware.get_iface_name(name)
 end
 
 function hardware.get_link_led()
-    local led = xpcall(
-        function() return "/sys/class/leds/" .. hardware.get_board().led.rssilow.sysfs end,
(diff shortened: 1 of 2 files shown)


aanon4:
Also, minor bug fixes where I was using xpcall incorrectly.

aanon4:
Any chance we can get this into the nightly?

dman776 (approved):

PR #216, opened 2022-02-11
Fix sysinfo link

Chris2fourlaw:
sysinfo.lua doesn't exist, correct path should just be sysinfo. Tested


aanon4 (approved):

aanon4:
Out of curiosity, did you stumble across this error reading the code or is there an unsupported device being incorrectly flagged?

Chris2fourlaw:
My unsupported device is being correctly flagged 😉

PR #220, opened 2022-02-16
Lua userpages: setup

aanon4:
The setup page is the first major "writable" page to test (and associated node-setup backend). It has a lot of configurations depending on the device so testing on various different devices and configuration is desirable. Because this page can modify the configuration state it could very easily **brick your node** and require physical access to fix ... but hopefully not.

The old setup page his renamed "setup.pl"


dman776 (approved):
looks good.

Issue #224, opened 2022-02-23
rssi_monitor.lua issue


aanon4:
Wed Feb 23 22:29:26 2022 daemon.err manager.lua[993]: /usr/bin/lua: /usr/local/bin/mgr/rssi_monitor.lua:60: attempt to concatenate a nil value
Wed Feb 23 22:29:26 2022 daemon.err manager.lua[993]: stack traceback:
Wed Feb 23 22:29:26 2022 daemon.err manager.lua[993]: 	/usr/local/bin/mgr/rssi_monitor.lua:60: in main chunk
Wed Feb 23 22:29:26 2022 daemon.err manager.lua[993]: 	[C]: in function 'require'
Wed Feb 23 22:29:26 2022 daemon.err manager.lua[993]: 	/usr/local/bin/manager.lua:65: in main chunk
Wed Feb 23 22:29:26 2022 daemon.err manager.lua[993]: 	[C]: ?

PR #225, opened 2022-02-23
Fix typo when checking for multiple antennas

aanon4:
https://github.com/aredn/aredn/issues/224


ae6xe (approved):
code review:  change needed to be absolute path.  looks good.

Issue #223, opened 2022-02-23
Problems with Nightly 899


k6ccc:
After seeing an issue reported with enabling RF via the lua setup page over on the Ventura county MatterMost, I did an update a GL-iNet USBB150, a GL-iNet AR750, a Mikrotik RB-LHG-5HPnD-XL, and a Mikrotik RB-LHG-2nD-XL from Nightly 887 to Nightly 899.  All were done via RF connections, and all reconnected to RF after the update.  So far so good.
I then connected a computer to the LAN port of the LHG-5HPnD and disabled the RF.  When the page advised that a reboot was needed, I could not reboot the node via the Reboot button.  That would result in a "Bad Gateway" error.  If I went back to the IP of the node, and then the setup page, it again wanted me to reboot.  Tried several times both with setup and setup.pl.  Finally found that I could go to the Administration page and successfully reboot the node.
After the reboot, sure enough, RF was off.  When I tried to turn it back on, I was able to enable RF, but again, any attempt to reboot the node from the setup or setup.pl pages resulted in the "Bad Gateway" error.  Again, I was able to reboot the node from Administration page.

aanon4:
Any chance you could attach the output from the support tool page just after you get the 'bad gateway' message?

http://<YOUR NODE NAME HERE>.local.mesh:8080/cgi-bin/supporttool

Alternative, if you log into the offending node, can you send the output of the 'logread' command immediately after getting the 'bad gateway' message.

Thanks

k6ccc:
The original reported problem from MatterMost was on both a hAP and a NanoStation M2.  That report from K1KY was that when trying to enable RF, he got a "Configuration NOT saved!".  Additionally it said: idconfig: bad address 'eth1' /usr/bin/lua: /usr/lib/lua/utils.lua:455: bad argument # 1 to 'tonumber' (string expected, got nil) stack traceback: [C]: in function 'tonumber' /usr/lib/lua/aredn/utils.lua:455: in function 'mac_to_ip' /usr/local/bin/node-setup:88: in main chunk [C]: ?
Hopefully I typed that right (it was originally screen capture).

k6ccc:
How can I post the support file?  girthub says it does not support that file type.

aanon4:
In that case just email it to me (tim.j.wilkinson@gmail.com)

k6ccc:
Sent via E-Mail

aanon4:
[...]

PR #228, opened 2022-02-24
Add the interface guessing code like there is in perl

aanon4:
Fix for https://github.com/aredn/aredn/issues/223


dman776 (approved):
looks good

PR #229, opened 2022-02-24
Move multi ant detection into main run method

aanon4:
This avoids some startup failures where we try to detect the antenna setup
too early in the node boot up process.


dman776 (approved):
looks good

PR #232, opened 2022-02-25
Lua version of fwinfo

TheRealIanJ:
Hi,

I'm new to lua but have a long history with perl. Hopefully I can help your effort to migrate to lua.

I see that some .pl files have been replaced with lua versions that have no file suffix. I have done the opposite with this file as the perl file has no suffix. I hope that's ok?

Best regards,

Ian

### DIFF
--- a/files/www/cgi-bin/fwinfo.lua
+++ b/files/www/cgi-bin/fwinfo.lua
@@ -1,0 +1,49 @@
+#!/usr/bin/lua
+--[[
+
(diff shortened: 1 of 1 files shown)


aanon4:
Hi,

Please check the following repo: https://github.com/kn6plv/aredn/tree/lua_1

This is where all the porting work has been done - we're just rolling it into the nightly slowly so we can manage the bugs better. I'm afraid there's already a port of fwinfo to lua there.

Thanks
Tim

Issue #231, opened 2022-02-24
Reboot command fails from Setup, but not Administration


Orv:
Under build 906, the Reboot button under Setup yields "Bad Gateway", but works as expected under Administration

Note this was observed on two separate Mikrotik hAP AC Lite routers, but NOT on a Mikrotik RBLHG-2nD.

Support Data will be emailed to KN6PLV

aanon4:
Please wait for the next build and re-test (some fixes only went in this morning).

Orv:
Will do.

aanon4:
Possibly a dup of https://github.com/aredn/aredn/issues/223 but currently can't and don't have a way to reproduce. More information greatly appreciated.

ab7pa:
This issue can be closed.

PR #236, opened 2022-02-28
Protect against OLSRD returning nil results

aanon4:
If we fail to retrieve information from OLSRD, make sure we don't fail, instead returning empty arrays (which seems like an appropriate default).

https://github.com/aredn/aredn/issues/231


aanon4:
Also https://github.com/aredn/aredn/issues/223

aanon4:
@dman776 Any chance on getting this one into tonight build?

dman776 (approved):

PR #242, opened 2022-03-01
Fixes for first boot setup

aanon4:
When a node is first flashed and setup is called, things are a little different which caused various setup issues (resulting in the dreaded "Bad Gateway" error). Fixed.


ae6xe (approved):
Code review;  looks good.

dman776 (approved):
nice one.

PR #245, opened 2022-03-02
Fix map when lat/lon unknown

aanon4:
Was incorrectly treating blank lat/lon as valid coordinates ("" is true in lua, but false in perl)


dman776 (approved):
good catch

PR #248, opened 2022-03-02
Only one device name per interface name (some devices have 2)

aanon4:
Apparently some devices have multiple device assigned to each interface ... so only return the first one.

### DIFF
--- a/files/usr/lib/lua/aredn/hardware.lua
+++ b/files/usr/lib/lua/aredn/hardware.lua
@@ -144,7 +144,7 @@ function hardware.get_iface_name(name)
         return "eth0.2"
     end
     -- Maybe the board knows
-    return hardware.get_board().network[name].ifname
+    return hardware.get_board().network[name].ifname:match("^(%S+)")
 end
 
 function hardware.get_link_led()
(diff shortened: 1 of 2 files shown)


ae6xe:
An include file is setup to add the bridge option for the interface definition when there are 2 listed.  I think when it needs both interfaces, this is pulled directly from board.json.   But in this case the bridge interface is “br-lan” format.  If the lua code is 1 to 1 translation of the Perl, this should work. Note this all changes when merging in openwrt upgrade branch.

aanon4:
This is part of the board.json file on older NSM2 devices:

```
"network": {

"lan": {
	"ifname": "eth0 eth1",
	"protocol": "static"
},
"wan": {
	"ifname": "eth0.1 eth1.1",
	"protocol": "dhcp"
},
"dtdlink": {
	"ifname": "eth0.2 eth1.2",
	"protocol": "static"
},
"wifi": {
	"ifname": "wlan0",
	"protocol": "static"
}

},
```

Unless I pick the first one I have situations where I do things like "ifconfig eth0 eth1" rather than "ifconfig eth0". This is just copying what the perl code is doing in node-setup.pl:

```chomp ($lanintf=`jsonfilter -e '@.network.lan.ifname' < /etc/board.json | cut -f1`);```

Note the "cut -f1" at the end to extract just the first device.

ae6xe:
I do recall doing this, where the interface is needed to find MAC addresses.  As long as there was always a 1 to 1 translation of the perl code, then this will still work.  Another place in the code, it needs both interfaces to be returned.  it was all a bit hackish...

Issue #249, opened 2022-03-02
NB942 issues


ab7pa:
Basic Settings page:
Active Settings box: changed distance setting and it saves/persists after reboots. Changed the Tx Power from dropdown list and it appears to save, but does not persist after reboot.  Always reverts to the max setting in the list.

ab7pa:
Also, on the Ports page: checking the "Do Not Propagate" checkbox and creating a DNS Alias are supposed to display different colored entries on the Mesh Status page.  Here is the original PR for that:
https://github.com/aredn/aredn_ar71xx/pull/569

ab7pa:
I also just noticed that all of the saved settings in the Advanced Configuration page are blank.  I had not noticed that happening after a SAVE --> REBOOT before.  Will verify that issue as well.

Yes, just confirmed that Set to Default followed by Save Setting on any row in Adv Config will blank all the rows of that table -- sometimes also resulting in a Bad Gateway message.

aanon4:
I opened this last comment as a new bug. Let's not add more to this one.

ab7pa:
This issue can be closed.

PR #250, opened 2022-03-02
tx power => txpower (typo)

aanon4:
Typo when setting power using iw command
https://github.com/aredn/aredn/issues/249

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -449,7 +449,7 @@ if (parms.button_apply or parms.button_save) and wifi_enable == "1" and phy then
     else
         os.execute("iw phy " .. phy .. " set distance " .. wifi_distance .. " >/dev/null 2>&1")
     end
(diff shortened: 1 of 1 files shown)


ab7pa:
@aanon4 Tim, I applied this fix but for some reason the new value does not persist after a reboot.  Defaults back to highest Tx Power in the dropdown.  Is this a UI issue or something in the backend config save process?

aanon4:
My observation with both the old Perl and new Lua was that hitting apply made the value stick until reboot but unless you hit save it would not persist across reboots even with the Perl code. Is this not what you see?

ab7pa:
No, I select a new value from the dropdown list, click APPLY, then click SAVE CHANGES, then click REBOOT.  After reboot it has reverted to the old value.

dman776 (approved):

aanon4:
Fixed in https://github.com/aredn/aredn/pull/252

PR #254, opened 2022-03-03
Fix poor dhcp and alias patterns.

aanon4:
Patterns were a little too loose and could match poor content causing failures


dman776 (approved):
Looks good

Issue #253, opened 2022-03-03
Saved settings in the top 16 rows of the Advanced Configuration page are blank.  I had not noticed that happening after a SAVE --> REBOOT before.


aanon4:
I also just noticed that all of the saved settings in the top 16 rows of the Advanced Configuration page are blank.  I had not noticed that happening after a SAVE --> REBOOT before.  Will verify that issue as well.

Yes, just confirmed that Set to Default followed by Save Setting on any of the upper 16 rows in Adv Config will blank all those rows of the table -- sometimes also resulting in a Bad Gateway message.

_Originally posted by @ab7pa in https://github.com/aredn/aredn/issues/249#issuecomment-1057494325_

aanon4:
@ab7pa Can you provide a way to reproduce this. I tried what you seem to describe without any issues:
1. Boot up node.
2. Go to Advanced Configuration Page. All boxes are currently populated with the correct information.
3. Hit "Set to Default" on the top box. Nothing changes.
4. Hit "Save" on top box.
5. Reboot
6. After reboot everything on the Advance Page is still okay.

If you could provide a similar set of steps which makes it fail, that'd be fantastic!

ab7pa:
Sent a screen capture video separately showing this behavior after a fresh reboot with NB942.

aanon4:
https://github.com/aredn/aredn/pull/255 (in part)

ab7pa:
This issue is still reproducible with NB956 this morning.

aanon4:
Sigh - but I'm not surprised.
Could I ask a favor - can you send me the system data dump again, but this time right before the finally reboot before you see the issue?

ab7pa:
Sent 3 support bundles separately, each at key steps in node life cycle.

ab7pa:
This issue can be closed.

dman776:
Looks like line 71 is the culprit

PR #257, opened 2022-03-04
Remove potentially problematic copying of /etc/config.mesh/aredn

aanon4:
Remove the potentially problematic 'filecopy("/etc/config/aredn", "/etc/config.mesh/aredn")'

/etc/config.mesh/aredn sometimes gets truncated, although reproducing it is difficult. This
is one of the few (only?) places this file is mindlessly written to, so instead use cursors to update it.

https://github.com/aredn/aredn/issues/253

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -50,6 +50,9 @@ local errors = {}
 local output = {}
 local hidden = {}
 
+-- uci cursor
+local cursor = uci.cursor()
+
 -- helpers start
 
 local rf_channel_map = {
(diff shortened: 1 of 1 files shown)


aanon4:
https://github.com/aredn/aredn/issues/253

aanon4:
I promise this is the last one today :-) @dman776 @ae6xe

ae6xe:
@aanon4 We're just trying to stay out of the way, to move all the high quality effort you are putting in forward!   You really don't need our review to get this stuff in!  I am humbled and in awe of the contribution you are making and can only hope that when I retire, I am able to do the same.  ...It will be another 5+ years before I retire :) .

ae6xe (approved):
looks good!

aanon4:
@ae6xe Thanks! I'm lucky enough to have the time, and good people to work with.

PR #266, opened 2022-03-06
Network fixes for older nodes using bridges

aanon4:
Well ... Joe @ae6xe  did warn me this was more complicated than I through ... so here we are.

Some nodes assign multiple ethernet devices to each "virtual" device (e.g. lan, wan, etc.) are much of this code didn't understand this. A earlier fix solved some of these issue, but didn't add all the ethernet to bridges correctly.

Also broken was the creation of the dtdlink which can also be a bridge, even if only one device is attached to it.

I suspect more subtleties will be discovered here, but for now this solves known problems with older NSM2/5 devices.

### DIFF
--- a/files/usr/lib/lua/aredn/hardware.lua
+++ b/files/usr/lib/lua/aredn/hardware.lua
@@ -128,7 +128,7 @@ function hardware.get_iface_name(name)
     end
     local intfname = cursor:get("network", name, "ifname")
     if intfname then
-        return intfname
+        return intfname:match("^(%S+)")
     end
     -- Now we guess
     if name == "lan" then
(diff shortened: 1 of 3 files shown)


ae6xe (approved):
Yes, the 1st two look correct.   To determine a mac address does not need both bridged interfaces, it just needs 1 interface.     the 3rd change...   There will never be a wireless interface setup as part of a bridge -- this needs to indeed be unique interfaces to uniquely route through the 2 paths and not add extra traffic on the 2 RF channels.   Thus, **row 115 would always have 1 entry to match** -- although the match doesn't hurt,   In the future we would have wifi1, wifi2, ... interfaces to find the names of, when support is added to configure multiple wireless cards to be mesh wireless interfaces.

aanon4:
@ae6xe @dman776 So I think we're okay with this (at least for the moment)? I don't have much older hardware to test on, but without these fixes the current nightly badly breaks my old NSM52 nodes (you have plug a cable into the secondary ethernet to recover)

aanon4:
Currently tested this on an old NSM2 (32M model), an AR150 and a new hAP and it appears good.

Issue #267, opened 2022-03-06
Reboot msg always shows LAN subnet changed


ab7pa:
The reboot message always seems to show the "LAN subnet changed..." message, even when no config change impacts the LAN network.  Wondering if the tests in the **reboot** function (lines 147 - 159) of **/www/cgi-bin/setup** may not be working as intended.  This screenshot was captured after decreasing Tx Power and saving/rebooting.  Clicking Reboot on the Adv Config page also still displays the LAN subnet message even without changing any settings.

![2022-03-06_07-42](https://user-images.githubusercontent.com/69524416/156928780-8737a10a-1f62-48ee-b64e-f5b3855ffe1d.png)

ab7pa:
This issue was fixed by PR#268.  Closing.

PR #268, opened 2022-03-06
Fix mis-reporting of changed LAN subnet on reboot

aanon4:
https://github.com/aredn/aredn/issues/267


ab7pa:
Confirmed that this PR fixes #267 on both pages where it occurred.  THX @aanon4 Tim!

dman776 (approved):
Confirmed by ab7pa

PR #270, opened 2022-03-07
Fix url format for no-link service names

aanon4:
Service names, which were not links, were being saved with a missing port number - breaks advertising them


ab7pa:
Perfect!  I confirmed that this fixes the issue we were seeing on the ports page.  THX

ae6xe (approved):
tested by @ab7pa

PR #275, opened 2022-03-09
Dont truncate /etc/config/vtun on tunnel install

aanon4:
Create /etc/config/vtun if it doesn't exist, but don't truncate it if it does.


dman776 (approved):

ab7pa:
Tested using global tunnel node and this fixes the issue.  THX Tim.

Issue #276, opened 2022-03-09
977 - Advertised Service error


k6ccc:
On my hAP (Nightly 977), I happened to be looking at the "Port forwarding..." page and it showed an error on the Advertised Service for the website that runs on an attached RasPi.  Confirmed that the webpage is working correctly, and nothing I did would clear the error.  Finally deleted the advertised service and rebuilt it.  Getting the same error.  I was on the page to attempt to set up a port forward, and could not save it because of this error. 
![Error-1](https://user-images.githubusercontent.com/36935831/157558637-d9c9bc54-c670-41b8-b8f4-8ac4f2c7ada8.png)

I have a downloaded support file, but unable to upload it here.

Jim

aanon4:
Could you retry with Nightly 991?

k6ccc:
I will attempt to have time this evening and let you know.

Jim




-----Original Message-----

ab7pa:
This was fixed in [PR 269](https://github.com/aredn/aredn/pull/269)

k6ccc:
Had a few minutes so updated.  The Advertised service was fixed, but the upgrade blew away all my tunnels.

aanon4:
That one will be fixed in tomorrow's nightly.

k6ccc:
Rebuilt all 13 tunnels manually - fortunately I have a spreadsheet with all the required information.  Two things of note:
1) The Tunnel Server Network IP range changed.
2) The Tunnel Server page does not show the tunnels active, however Mesh Status does show them active, and the far ends (on the could I could easily check) show the tunnels active on their respective Tunnel Client pages.

aanon4:
Re: 1 - do you know what it was and what it is now?

k6ccc:
Just noticed that with 991, Advertised services are not showing up at all on a Mesh Status - either locally or on a different node.
This was observed on a hAP that has an advertised service of a web page that is running on a RasPi, and also an IP phone.  Noticed when I tried to add something as an advertised service, it did not show up - and neither did the pre-existing web page or IP phone.  Created another advertised service and it also did not appear on that hAP, or another node.  Tried it on my other hAP with 977 and the advertised service shows up fine
![991 advrtised services error](https://user-images.githubusercontent.com/36935831/157810069-572cdaac-4145-4def-9e75-00ebf57eded3.png)

aanon4:
Is fixed in the Friday build.

PR #277, opened 2022-03-10
Fix active tunnel indicator for tunnel server

aanon4:
Tunnel server was working but  check for displaying active status was using the wrong value.


aanon4:
https://github.com/aredn/aredn/issues/276

dman776 (approved):

Issue #272, opened 2022-03-08
Remove cron


aanon4:
The only thing cron does now is update aredn messages every 12 hours. Absorb this functionality into manager.lua and remove cron to save memory.

ab7pa:
Cron does allow you to set the AREDN message refresh rate to any interval you desire.  If integrated into manager.lua there should still be a way for node owners to update their message refresh rate, maybe as an Advanced Configuration setting?

aanon4:
There's no UI to change this setting currently? You just edit the file by hand?

PR #279, opened 2022-03-10
Cron is only running to poll AREDN messages, so remove it.

aanon4:
Move polling of AREDN message into the Lua Manager.

This is not on the critical path for dumping perl, but as this was the only thing left in cron, it's a "nice to have" from reducing what's running on the system (and consuming RAM). Semantically its not quite the same as what it replaces: the old cron would poll at midnight and midday, while this will poll every 12 hours from boot.

I added a config option (per @ab7pa suggestion)


aanon4:
https://github.com/aredn/aredn/issues/272

ae6xe (approved):
looks good.

dman776 (approved):
Code looks good

PR #280, opened 2022-03-10
Fix the pattern used to validate services

aanon4:
Broken service pattern

### DIFF
--- a/files/usr/local/bin/olsrd-config
+++ b/files/usr/local/bin/olsrd-config
@@ -131,7 +131,7 @@ end
 if nixio.fs.stat("/etc/config/services") then
     for line in io.lines("/etc/config/services")
     do
-        if line:match("^%w+://[%w%-%.]+:%d+(/[^|]*)?|[tu][cd]p|%w") then
+        if line:match("^%w+://[%w%-%.]+:%d+.*|tcp|[^|]+$") or line:match("^%w+://[%w%-%.]+:%d+.*|udp|[^|]+$")
             services[#services + 1] = line
         end
     end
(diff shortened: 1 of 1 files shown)


ab7pa:
With this olsrd-config edit the /etc/config/services file looks good, but the /var/run/services_olsr file no longer contains any of the local services.  Therefore, the Mesh Status page does not display the services for this node under the Local Hosts section where there should be 11 services listed.
Screen Shot 2022-03-10 at 10 07 20 AM

aanon4:
I think (need more coffee still) that /var/run/services_olsr is generated by OLSR from the information it receives on the network. If services are missing then there's probably an issue with olsrd-config still. Can you post the output from olsrd-config and the old olsrd-config.pl so?

ab7pa:
Files attached here.  Diff shows that **PlParam** values are empty in the lua version but are populated with the local host in the Perl version.
[olsrd-config-lua.txt](https://github.com/aredn/aredn/files/8225769/olsrd-config-lua.txt)
[olsrd-config-pl.txt](https://github.com/aredn/aredn/files/8225770/olsrd-config-pl.txt)

aanon4:
@ab7pa thanks - can you try the new fix?

ab7pa:
Yes, that fixed the issue.  Made the edit and restarted OLSR and everything came back up correctly.  Thanks very much!

aanon4:
Excellent catch as always - much appreciated

aanon4:
@dman776 @ae6xe I think this is good to go now

ae6xe (approved):
tested by @ab7pa .

PR #287, opened 2022-03-12
Default phy0 if we fail to read one from the system

aanon4:


ae6xe:
Every device AREDN images are built for has a phy0.   If this condition were to occur, then there is a fundamental issue with the hardware or other root cause that would need to be resolved to have a functioning system.    It may be a better option for this condition,  should it ever occur, is to set the interface as if "mesh RF" is not enabled.  (setting this value too.)   In this case the interface would be eth0.3975 (from memory).   did this situation come up in testing?

aanon4:
Yes, it was someone testing with WiFi mesh off, so it attempted to get the physical wifi device for eth0.3975 which gave it back a 'nil'. Defaulting to 'phy0' when this happened seemed like a good fallback.

ae6xe:
With mesh RF off, any functions that attempt to do wireless actions should really  be turned off.  This is incomplete in the Perl code, eg errors are occurring to set the xmit  power probably, but no one sees the errors.   If we know where this method was called from,  jump around it if mesh RF is off.   I think we may already have the WiFi scan option not showing if  mesh RF is off.  The error may just get pushed elsewhere to still show up in lua with this change. Will the xmit power set code error out in lua since phy0 is not configured?

More work may need to be done for AC device support.  Ubiquity AC devices all have 2 wireless cards phy0 and phy1.   We have to make sure the right interface is called everywhere depending on how the cards are configured.  Any place that hard codes phy0 now needs to call get_interface as well.

aanon4:
@ae6xe Here's an alternative. Instead of defaulting phy I just avoid the one place it's used.
(Currently this page is giving "Bad Gateway" when mesh rf is disabled)

ae6xe (approved):
good option.  Looks good.

PR #289, opened 2022-03-13
Quote escapes (take 2)

aanon4:
Copy over the perl escapes for the node descriptions and contact information.
The perl code explicitly escapes certain fields and this was not reflected in the lua code - now it is.

Also, fixed a couple of typos.


ae6xe (approved):
looks good.

PR #292, opened 2022-03-14
Remove xinetd (which has little to do)

aanon4:
xinetd was only being used by the system to proxy network calls to the local olsrd dotdraw plugin. If we change the olsrd config options we can let olsrd handle this traffic itself. This frees up a little more ram (always good) and a little more flash.

It's not clear why this proxy was put in place as olsrd as olsrd behaves the same regardless of whether the request is locally generated or comes from the network.

@dman776 has suggested there may have been a concern that reading this data blocked olsrd itself. Looking at the code for the plugin it appears to correctly use select to make sure it can write to the socket before doing any writes. Also, the payloads themselves are quite small compared to the default tcp buffer size (~200K) so it would be unlikely to block even if it wasn't using select.

Thoughts?


ae6xe:
I made changes a few years back that resolved an issue with  OLSR live locking.   This had to do with the buffering of data sent/received and the right read/write calls taking place -- check out these patches in the aredn_packages repo.    I think xinetd was shielding this olsr plug in from doing the same.    We need to review what was done in the past to confirm we won't be opening up olsr to a live-lock situation.

ae6xe:
see patches:

[net/olsrd/patches/007-fix-do-not-block-when-sending-data-windows.patch](https://github.com/aredn/aredn_packages/commit/a925db8da25b71e34f9eada269b209d7ba6cc812#diff-a0e18519d0e634c728a0303bd175625845aba94f4cd0eb3628fe6f1353e15bdd)

[net/olsrd/patches/006-do-no-block-when-sending-data.patch](https://github.com/aredn/aredn_packages/commit/a925db8da25b71e34f9eada269b209d7ba6cc812#diff-8b16c31b27a949ef3a3c66bef6602211c524cb7acfac2649bd7cfe57d8bd8bbb)

aanon4:
Good to know! I can probably write a tiny change to the current code which uses FIONSPACE to limit how much is written to what would fit in the buffer, so the send call won't block (the code already handles partial buffers being sent).

aanon4:
Unless I'm missing something (again), the change to the dot_draw plugin is super small. Here's the snippet of sending code in the olsrd plugin:
```
  result = select(outbuffer_socket + 1, NULL, &set, NULL, &tv);
  if (result <= 0) {
    return;
  }

  if (FD_ISSET(outbuffer_socket, &set)) {
    result = send(outbuffer_socket, outbuffer.buf, outbuffer.len, MSG_DONTWAIT /* <-- FLAG ADDED HERE */);
    if (result > 0)
      abuf_pull(&outbuffer, result);

    if (result < 0) {
      /* close this socket and cleanup*/
```
It's literally just adding that flag to the send. My reasoning is: the select at the top makes sure you can write at least one byte to the socket, the flag make sure the send won't block but write as much as it can, the length of which is returned, and the remaining code already handles partially sent packets.

ae6xe:
That looks consistent with the prior changes.    The risk-impact of a lingering issue is extremely low to make this change and remove xinetd.   The other plugins were called frequently and by some of the UI pages as I recall.  dotdraw is only called when a user runs a perl script -- which is never on most nodes.    I vote we go with this, including the MSG_DONTWAIT update.   The plugin config would need a port change, to be what xinetd is configured for now.   The firewall rules should still work allowing the incoming on this port direct to the plugin.

aanon4:
https://github.com/aredn/aredn_packages/pull/8

aanon4:
This PR already contains the necessary migration stuff to remove xinetd and change the port configuration on old nodes, and setup new nodes correctly.

ae6xe (changes requested):
The permanent change to the olsrd config file for port 2003->2004 conversion is in /etc/config.mesh/olsrd.    The 95_aredn_olsrd_change_dotdraw script should change in both /etc/config and /etc/config.mesh (or just /etc/config.mesh and save setup).

aanon4:
As far as I can tell, /etc/config.mesh/olsrd is replaced during the upgrade process - it does not appear in /etc/ardnsysupgrade.conf and so isn't preserved. The new, updated, /etc/config.mesh/olsrd will replace it

ae6xe:
with a sysupgrade with keep settings occurs, a save setup occurs, so /etc/config/olsrd will be updated with the olsrd template and already have the updated port #.

aanon4:
You're right. I've removed the upgrade script.

ae6xe (approved):
looks good.

PR #293, opened 2022-03-15
Fix typo in iperf fw restart patch.

aanon4:
The iperf3 package installs a firewall rule, but the script to restart
the firewall on install had a typo meaning it wasn't called. Instead the
new firewall setup required and otherwise unnecessary reboot.


ae6xe (approved):
good catch.

PR #294, opened 2022-03-16
Improve the firmware upgrade process

aanon4:
The old firmware upgrade process attempted to free up RAM by reusing
the 'upgrade_kill_prep' script (which is used internally by '/sbin/sysupgrade').
Unfortuantely this doesn't work as intented. While the script will go about
killing various services, 'procd' just goes and starts them up again using
quite a bit more memory in the process. Instead this script just kills
the various daemons '-9, no questions asked' and then runs the associated
'/etc/init.d/xxx stop' script to instruct 'procd' not the start them up again.
This gets us to the place the original script was trying to go.

Also, an extra syntax fix in '007' patch (need spaces around the [ .. ]) as it was erroring

### DIFF
--- a/files/usr/local/bin/upgrade_prepare.sh
+++ b/files/usr/local/bin/upgrade_prepare.sh
@@ -1,0 +1,61 @@
+#! /bin/sh
+true <<'LICENSE'
+  Part of AREDN -- Used for creating Amateur Radio Emergency Data Networks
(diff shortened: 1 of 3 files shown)


aanon4:
Now clearing away service as soon as we detect a firmware upgrade is starting, so we can get space in temp before we start downloading.

dman776 (approved):

PR #295, opened 2022-03-16
Add /var/run/services_olsr to support tool data

aanon4:
Had to request this file by-hand during the debugging process for some nodes, so add it by default.

This file wasn't included in the old perl version.


ae6xe (approved):
Looks good.

PR #297, opened 2022-03-16
Modify the style sheet URL to let the browser cache

aanon4:
Modify the style sheet URL to let the browser cache while still supporting switching the style.
He old code appended the time which made it simple to switch styles on the node, but forced everyone to reload the stylesheet every time. Now append the actual style after the ? in the url, so we cache the style and let it be switched.


dman776 (approved):
nice change

PR #300, opened 2022-03-17
Automatically update package dependencies if package install fails

aanon4:
When uploading a package by hand which depends on an uninstalled system package, unless the packages have been updated since boot, the install will fail. This can be particularly confusing because we persist the downloadable package list across reboots but opkg itself does not keep its internal dependency list. Instead detect the dependency failure, update the package list, and try again.


dman776 (approved):
nice improvement

PR #301, opened 2022-03-17
Dont truncate aliases on upgrade

aanon4:
Create file it it doesn't exist, but don't truncate it.


dman776 (approved):
good catch

aanon4:
Credit to @ab7pa for sending me the issue

r1de:
agreed. good catch. :man_facepalming:

PR #302, opened 2022-03-17
Include tunnels by default

aanon4:
We discussed including tunnels in node by default, including the ability to upgrade with active tunnels. These are the changes to make that happen.

I'm putting this up as a draft in case anyone wants to try it out. I've done some basic migration and configuration testing and everything seems in order, but as a non-tunnel user I could easily have missed many things I'm unfamiliar with.

On my test node (an old 32M nanostation) flash usage increases by 184k (/dev/root goes up but /overlay goes down). Ram usage when not actively using tunnels is the same. Download image increases by 128K


dman776 (commented):
i think there is code that can be removed in sysinfo.json (around line 103).   We don't need 'tunnel_installed' anymore, but, do want 'active_tunnel_count'
be sure and increment the version number from 1.9.

dman776 (commented):
looks good to me.

dman776 (approved):
Been waiting a long time for this.   Good changes.

PR #310, opened 2022-03-19
advancedconfig: tun wan only

aanon4:
Add an option (for firewall settings in the vtund app) to block/unblock tunnel traffic from going across the mesh. You probably never want tunnel traffic across the mesh, and this can happen accidentally if a tunnel node's WAN network is down.


ae6xe (approved):
Looks good.

mathisono:
About time someone noticed this as something to stop under specific
conditions!

73 de kj6dzb

aanon4:
@mathisono I've seen you mention this and subsequently has been on my list for a while

PR #315, opened 2022-03-21
reduce slider chunkiness

ab7pa:


aanon4:
Any chance we could align the text with the smaller switch?

PR #317, opened 2022-03-23
Add periodic tasks in the style of cron hourly, daily and weekly scripts

aanon4:
Allows user and/or packages to install cron scripts which will be run approximately every hour, day, or week.

A use for something like this came up while I was looking at the mesh ntpd stuff and I really wanted to a way to
refresh the peer list every now and again. Obviously having a daemon hanging around just to do this is very wasteful.
This is a simple alternative which is already familiar to Debian developers. Script are installed in the relevant /etc/cron.X directory and run on the appropriate schedule.


ab7pa:
Looks good.

dman776 (approved):
Looks good.   Glad to have this capability back

Issue #318, opened 2022-03-24
Discussion: ntpd on a node


aanon4:
(re: https://github.com/aredn/aredn/issues/221)

Rather than opening a PR with a possible solution, I thought I'd highlight some options for discussion first as there are several:

1. **By default, no ntp on a node.** Given that running ntp at all takes resources, perhaps we just shouldn't run it at all on nodes by default. After all, day-to-day, most nodes don't need the time set particularly.

2. **Run ntp on startup.** This is what we currently have. Ntpd runs on boot to set the initial time on the node and then assume the node will keep good-enough time afterwards.

3. **Run ntp periodically.** A slightly improved version of (2) given that, actually, nodes probably don't keep great time over long periods. Perhaps reset the time every day.

4. **Run ntp.** Here we run ntpd all the time, so a node will keep accurate time over long periods.

Orthogonal to the above is how a node chooses the time server to sync with. Currently we have a default set one on the main page which may, or may not, be reachable. It can be changed, although I have no idea how many people do that. An alternative to that is for a node to search out NTP server on the local mesh and use those for their time sync. Auto-configuration is nice, but it's not something we do anywhere else on the node at the moment. Also, it's a bit of a heuristic process as there's no well defined way to advertise ntp servers on the mesh (for now, it's basically looking for "ntp" in service names and then seeing if they're really ntp servers).

Thoughts?

K6AH:
I think I'm landing on #3 using the default NTP server in the config, and on failure search via auto config as described.  Perhaps randomize when it gets the time from the mesh so they all don't do so at the same time.  The trade off of course is memory, and you don't say how much space this will consume, but I can't image it's much.

aanon4:
@K6AH A running ntpd server uses about 1M (at least that's the footprint I see on my AR750)

ab7pa:
The legacy ntpclient uses about the same amount of memory as ntpd (932kb).

wu2s:
I prefer #3. Just run periodically - my preference is about 4 times daily. That should be more than enough to keep reasonably in sync with correct time.

K6AH:
Wow, that seams like a lot memory.  So, between adding the tunnel and ntp features we've consumed all we saved by removing perl?

aanon4:
[...]

PR #319, opened 2022-03-25
Run NTPD daily to sync time against network clock

aanon4:
Here's a new NTPD proposal (requires the change in PR https://github.com/aredn/aredn/pull/317)


aanon4:
https://github.com/aredn/aredn/issues/318

ab7pa:
The first script to use the periodic feature.  Looks good.

files/etc/mesh-release
```diff
@@ -1 +1 @@
-3.1.1TUNNEL
```
dman776, line 1: Drop this file

dman776 (changes requested):

files/etc/mesh-release
aanon4: Oops. Will fix it later this evening.

aanon4: Fixed

dman776 (commented):
I like it.   @ae6xe any feedback

dman776 (approved):

PR #321, opened 2022-03-30
Fix bad gateway when switching to LAN NAT mode

aanon4:
Missing hidden variable caused Bad Gateway when switching to NAT mode


dman776 (approved):
good catch

PR #325, opened 2022-04-02
Fix displaying "Previous neighbors" with empty hostnames

aanon4:
Neighbors names can be missing and we were just displaying blanks when this happened. Instead we now display IP addresses (as we should have originally).
Also fixed a few other bugs related to the upkeep of this information.


ae6xe (approved):
looks good.

PR #326, opened 2022-04-05
Optimize getRemoteNodes

aanon4:
Remove the nested loops O(n^2) and replace with linear loops and hash tables O(2n).


dman776 (approved):
Much better

Issue #51, opened 2018-08-11
User configurable VLAN ID for WAN


kd2hkb:
It may not be an issue for all, or even most, but VLAN1 is locked by some manufacturers (Ubiquiti's Unifi line, for instance) which AREDN also uses for WAN. The ability for users to set their own VLAN IDs would resolve this & help avoid conflicts with other existing networks.

PR #329, opened 2022-04-16
Support changing the VLAN tag on WAN

aanon4:
Support changing the default VLAN tag for the WAN interface.
This comes up a bunch for people running complex networks (I had to buy a switch I use to specifically rewrite the VLAN tag of WANs) and providing this as an advance feature seems useful.
Also, added in a 'reboot' message for changing advance config options which need a reboot.

### DIFF
--- a/files/usr/local/bin/aredn_postupgrade
+++ b/files/usr/local/bin/aredn_postupgrade
@@ -107,7 +107,7 @@ do
 end
 
 -- specific settings for variables that are not in the default config but are added by the system
-for _, key in ipairs({ 'dmz_dhcp_end', 'dmz_dhcp_limit', 'dmz_dhcp_start', 'dmz_lan_ip', 'dmz_lan_mask', 'wifi_rxant', 'wifi_txant', 'wan_gw', 'wan_ip', 'wan_mask' })
+for _, key in ipairs({ 'dmz_dhcp_end', 'dmz_dhcp_limit', 'dmz_dhcp_start', 'dmz_lan_ip', 'dmz_lan_mask', 'wifi_rxant', 'wifi_txant', 'wan_gw', 'wan_ip', 'wan_mask', 'wan_intf' })
 do
     local v = cfg[key]
     if v then
--- a/files/usr/local/bin/node-setup
+++ b/files/usr/local/bin/node-setup
@@ -90,7 +90,7 @@ local dtdmac = mac_to_ip(aredn.hardware.get_interface_mac(lanintf:match("^(%S+)"
 local deleteme = {}
 local cfg = {
     lan_intf = lanintf,
-    wan_intf = "dummy",
+    wan_intf = aredn.hardware.get_board().network.wan.ifname,
     dtdlink_intf = aredn.hardware.get_bridge_iface_names('dtdlink')
 }
 
--- a/files/usr/local/bin/wifi-setup
+++ b/files/usr/local/bin/wifi-setup
@@ -55,14 +55,6 @@ done < $configfile
 meshif="$(uci -q get network.wifi.ifname)"
 meshphy="phy${meshif#wlan}"
 
-# set physical wan interface in network
-
-wan_intf=`cat /etc/board.json|jsonfilter -e '@.network.wan.ifname'`
-uci -c ${dropdir} -q batch > /dev/null <<-EOF
-set network.wan.ifname="$wan_intf"
-EOF
-uci -c ${dropdir} -q commit network
-
 rm -f "${dropdir}/wireless" 
 touch "${dropdir}/wireless"
 
--- a/files/www/cgi-bin/advancedconfig
+++ b/files/www/cgi-bin/advancedconfig
@@ -212,12 +212,13 @@ local settings = {
         default = "1000"
     },
     {
-        key = "network.wan.ifname",
+        key = "aredn.wan.vlanid",
         type = "string",
-        desc = "Specify WAN interface VLAN #. MUST be entered as ethX.YY where X is the interface (usually '0') and YY is the VLAN # to use.",
+        desc = "Specify WAN VLAN #",
         default = "",
         condition = "supportsVLANChange()",
-        postcallback = "postChangeWANVLAN()",
+        current = "currentWANVLAN()",
+        postcallback = "changeWANVLAN()",
         needreboot = true
     },
     {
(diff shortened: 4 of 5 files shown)


r1de:
:+1: :+1: :+1: :+1: :+1:

_years_ in the making.

ae6xe (changes requested):
There are some devices where the wan port is on a switch chip, e.g. wbs510-v2 .   This means files that are initialized on firstboot are created with this information to update the switch settings.   See the files /etc/aredn_include/swconfig.   Also, in openwrt upgrade branch, I had added the wan port to be on 2 different ports (added vlan 1 to the dtdlink vlan 2 port on the hap ac lite).   This way a user could use 1 cat5 cable to have wan and dtdlink, instead of 2 cables on 2 ports.

ae6xe:
After applying patches, look at the files linux/ath79/base-files/etc/board.d/02_network, and target/linux/ar71xx/base-files/etc/board.d/02_network.   If you see this notation, a device that uses the swconfig for setting vlan 1 is used:

 ucidef_add_switch "switch0" \
                        "0@eth0" "2:lan" "3:lan" "4:lan" "1:wan" "5t:dtdlink"

the order in this file determines the vlan used.  lan is first so gets a '0',  then wan, so gets a '1', then "dtdlink", so gets a 2 auto assigned.  

To change this assignment on a live node, the detail in /etc/board.json needs updated, and first boot scripts create the swconfig files, which then is included in /etc/config/network on save settings.

aanon4:
As a way to move this forwards we could not support this feature on devices which contain switches?

aanon4:
The incoming multiple-wan port change should be fine with this change I think. I mean, it totally neutral if you don't use it, and my old NSM5 already has multiple wan ports and this change was fine with that.

ae6xe:
These are the affected devices

**ar71xx images:**
cpe210
cpe220-v2
cpe220-v3
cpe510
wbs210
wbs510
gl-ar750s
nanostation-m-xw 
**ath79 images:**
gl-ar750s

I'd put out for others to weigh in on excluding these.    Best to detect if there is a vlan 1 in the swconfig file or board.json to hide the option, in case my list has an error.   I wouldn't want to hold hostage the devices this commit will work on to have this feature for the above devices -- it is incremental value delivered.

aanon4:
I went with simply not including the feature if there was *anything* in the swconfig file, which seemed like the safest option.

ae6xe:
There are several devices, e.g. the hap ac lite, where there is a switch for eth0 - lan and dtdlink.  But the wan is on eth1.   swconfig will be populated, but does not contain vlan1 and the code on this commit should work.

ae6xe:
check this code doesn't change the setting on the hap ac lite, GL-AR150, GL-AR300M16, and GL-AR750.  On these devices, the WAN port is "untagged", and the config settings should not change.

aanon4:
I think we should let the hAP go for now also - start with what will definitely work and add to it later.
I have checked the GL-AR150 (my usual debugging box) and it correctly switches from untagged to tagged traffic as I'd expect when setting a tagged device for WAN.

r1de:
I can possibly try to test this on one of our easily accessible "other vlan" devices tomorrow, just have to see how the day goes. 🤞 
From talking with Tim though, I think it'll be just fine.

aanon4:
More thoughts on hAP and switches: Once switches get involved there are lots and lots of configuration options available to the avid tinkerer. Even with the hAP there's no reason someone couldn't assign the WAN to a VLAN on the switch even though that's not what happens today.

So how about this approach. For "simple" (ie. no switch) devices we can use this change which gives easily changeable VLANs on many devices (where what is happening is obvious to the modest hacker). For people who really want to change things up, we do a follow-on which doesn't attempt to add UI for complex changes, but instead finds a way to persist these changes across upgrades (unlike today). Thoughts?

ae6xe:
Huge limitation with the vlan implementation.   A given port can not share untagged  with tagged.

aanon4:
> Huge limitation with the vlan implementation. A given port can not share untagged with tagged.

I assume you mean a switch port, yes? If we continue to exclude devices with switch chips in this PR, that shouldn't matter (I think).

ae6xe:
Yes, Only relevant if there becomes more options for end users to do custom switch port configurations.

aanon4:
Having slept on this a bit I think I'm going to tweak how the UI presents itself slightly. Rather than letting you specific the device to use (which is easy to implement) I'm going to switch to letting you specify the VLAN ID to use. This should make adapting this change for switch-enabled devices more sane.

aanon4:
Okay - same supported devices as before (so still nothing with a switch) but now specify just the VLAN ID for the WAN. This will make propagating the changes to devices with switches easier to do.  The original PR let you not only change VLAN ID but also the device itself which was undesirable, even in an advance option.

aanon4:
@r1de any chance you can give this a whirl before tomorrow evening?

r1de:
I will try, not getting home till about 6pm tho.

r1de:
Tested on one of our local customized devices and WAN interface VLAN ID switching works.
Went from VLAN 1 to VLAN 15. :+1: 
But, hold up, don't celebrate just yet... I re-flashed it, _with the same code_, just now, and it seems to have _not_ retained the VLAN ID change.
More on this once I can get back into that device. On our network if VLAN 15 goes away, the network breaks apart, and of course, I am on the wrong side.

aanon4:
There are various places the code sanity checks _setup and I missed one in the upgrade path - fixed now.

r1de:
Re-tested with new commit and VLAN ID is now retained across upgrades.
Nice work!!  :+1: :tada:

ae6xe:
Is this ready to commit to the repo?   Which devices did this end up to being applicable to (or not applicable to)?

aanon4:
Yes. We've finished testing it and it looks good.

It is limited to devices without a switch chip. I can build a complete list of supported/unsupported devices for the final release notes.

ae6xe (approved):
tested by Eric r1de.

dman776:
I think we need to discuss.   Can we do this for all devices?

dman776:
resolves #51

aanon4:
This doesn't do it for all devices, but yes we can do this for all devices. Those currently unsupported all contain switches, which is trickier, and I don't have enough of them (yet) to test that work. Despite that I think its valuable to support what we can now.

Somewhat randomly this is the first thing I ever wanted to change on my AREDN nodes as VLAN-1 is a no-go on my home network. Ended up buying a switch cable of re-writing VLAN tags to deal with the issue. This is cheaper :-)

dman776:
Need to update README.md to indicate that vlan 1 is the WAN port (by default, but, can be changed in Adv Config)

dman776:
any changes needed to /usr/lib/lua/aredn/hardware.lua - get_iface_name()
updates also needed to /www/help.html (WAN Wifi Client section references vlan1)

aanon4:
/usr/lib/lua/aredn/hardware.lua - get_iface_name() is good as it is.

aanon4:
I'll annoy @ab7pa about the best way to update the help information and documenting the adv option.

dman776:
we need that also, but, the readme.md and help.html are in this repo and should be clarified.

dman776:
should we change that the vlan input is numeric, between 1 and 4094, and not 2 (dtd)??

r1de:
I should add that this tested on a _live_ network.
and +1 to what @dman776 just said ^^ block VLAN2.

aanon4:
I decided that the very big scary red warning at the top of this page was sufficient. I don't think there's validation on any other setting here.

aanon4:
@ae6xe Yes, but I don't have enough devices with switches, so once some I get them I'll work on the changes.

PR #332, opened 2022-04-19
Add port dump to supporttool data

aanon4:
It's seems that OLSRD doesn't always open all the appropriate ports, so it would be good to see when this happens. Not a fix, but helpful.


ae6xe (approved):
looks good.

PR #336, opened 2022-04-20
Include iperf3 by default

aanon4:
I believe @K6AH asked if we could include iperf3 by default.


K6AH:
I believe there are at least 2 compelling reasons to include iperf:
1) We would expect competent network operators/admins to worry about choke points while optimizing their network.  They would use iperf in that management process.
2) It's a cool-enough tool it would motivate node owners to upgrade all of their nodes to the new UI right away.

It's also small enough to not severely impact available memory.  Only down side could be network impact, so we should discuss constraining these tests to short duration.

aanon4:
Importantly, this is just the iperf3 application, and doesn't come with any UI pieces (like that found in the iperfspeed package which sits on top of iperf3) nor does it run in server mode automatically. For now an operator would have to ssh/telnet into both nodes to run this by hand.

ab7pa:
Looks good.

K6AH:
Presumably it could be launched from within the UI "Tools" page.  Perhaps the server side could always be listening for a request.

ae6xe (approved):
looks good.

PR #337, opened 2022-04-21
Run the iperf3 server by default [For discussion]

aanon4:
Putting this up here for discussion. This change starts the iperf3 server on every node by default (rather than it just being installed).

Positive:
Having the running on every node makes it much easier to debug problematic network links and flakey nodes. Few people have iperfspeed installed (a UI for iperf3) making testing and finding bottlenecks on the network nearly impossible.

Negative:
Will people find it objectionable to have this running by default?

From personal experience, not having tools like this widely available on a network makes it very difficult to generate good data for fixing things. The SF mesh regularly goes through an identity crisis where people get frustrated with performance, threaten to burn everything down, but ultimately have no tools to make informed and useful choices. Things like this would help.

### DIFF
--- a/files/etc/init.d/iperf3
+++ b/files/etc/init.d/iperf3
@@ -1,0 +1,15 @@
+#!/bin/sh /etc/rc.common
+
+USE_PROCD=1
+START=91
+
+start_service() {
+    procd_open_instance iperf3
+    procd_set_param command /usr/bin/iperf3 -s -D
+    procd_set_param respawn ${respawn_threshold:-3600} ${respawn_timeout:-5} ${respawn_retry:-5}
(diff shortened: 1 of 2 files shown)


dman776:
What is memory footprint when running the server?

aanon4:
Just put up an alternative approach (https://github.com/aredn/aredn/pull/338)

ab7pa (commented):
I would vote for PR#338 personally.  Many aging nodes on our local mesh here in AZ simply do not have enough memory to run `iperf3 -s -D` all the time.

K6AH:
I think Steve has it right... although it wouldn't hurt to offer an incentive to upgrade some of this old equipment.

PR #339, opened 2022-04-22
Old version - now in mgr/snrlog.lua

aanon4:
Left over copy of "snrlog" which was migrated into the manager a while ago


dman776:
Good catch

Issue #116, opened 2021-05-31
3 GHz Nodes: API reporting wrong channel and freq.


FletchFFletch9:
When polling a 3GHz node it returns that it is using channel 3400 and freq 5.400 GHz. The channel (in this case) should be 80 and the freq should be 3.400 GHz. 

Robert
KC6RKG

http://w6jpl-rm3-pleasantspk-p2p-jpl:8080/cgi-bin/sysinfo.json?link_info=1

{
   "lon": "-117.623617",
   "sysinfo": {
     "uptime": "6 days, 17:16:34",
     "loads": [
       0.27000000000000002,
       0.23000000000000001,
       0.23999999999999999
     ]
   },
   "interfaces": [
     {
       "mac": "68:72:51:85:4E:EB",
       "name": "eth0.2",
       "ip": "10.133.78.235"
     },
     {
       "mac": "68:72:51:84:4E:EB",
       "name": "wlan0",
       "ip": "10.132.78.235"
     },
     {
       "mac": "68:72:51:85:4E:EB",
       "name": "br-lan",
       "ip": "10.34.119.89"
     },
     {
       "name": "tunl0",
       "mac": "00:00:00:00"
     },
     {
       "name": "eth0",
       "mac": "68:72:51:85:4E:EB"
     },
     {
       "name": "eth0.1",
       "mac": "68:72:51:85:4E:EB"
     }
   ],
   "api_version": "1.8",
   "lat": "33.797861",
   "meshrf": {
     "ssid": "AREDN-10-v3",
     "channel": "3400",
     "status": "on",
     "freq": "5.400 GHz",
     "chanbw": "10"
   },
   "tunnels": {
     "active_tunnel_count": "0",
     "tunnel_installed": false
   },
   "link_info": {
     "10.8.221.154": {
       "neighborLinkQuality": 1,
       "noise": -95,
       "linkQuality": 0.92100000000000004,
       "linkType": "RF",
       "hostname": "KC6RKG-NG3-SE5-SM33-RKG2PP.local.mesh",
       "tx_rate": 78,
       "olsrInterface": "wlan0",
       "rx_rate": 52,
       "signal": -74
     },
     "10.15.32.134": {
       "neighborLinkQuality": 1,
       "linkQuality": 1,
       "hostname": "AE6XE-PleasantsPk-RM3.local.mesh",
       "olsrInterface": "eth0.2",
       "linkType": "DTD"
     },
     "10.146.164.14": {
[...]

PR #345, opened 2022-04-27
Correct the frequency for 3GHz radio

aanon4:
I don't have a 3GHz radio to test this, so if someone could that'd be lovely, but it uses the same logic as getChannel()

https://github.com/aredn/aredn/issues/116


Orv:
I can test it.

Orv:
Looks good.   I think we can close this one.

PR #356, opened 2022-05-06
add ifname to api

ab7pa:


ae6xe (commented):
Also a dtdlink interface (and tunnel interfaces).    Maybe optional to add, so I'd put it out for others to comment if this should be added.  Also, wondering the use cases for someone to have this information.   1) snmp to see volume of data in/out these interfaces; 2) ?

ab7pa:
I was thinking that the newUI might be able to make use of the ifname -- for example to determine whether the WAN port requires VLAN tags, etc.  If this is unnecessary then we can close this PR.

ae6xe:
For the new UI, all the information in board.json is needed for setup.
Also, all the information of the current /etc/config files is needed.
Based on the default hardware definitions (board.json) and the current
usage of this hardware (/etc/config), then the UI will have the information
necessary to present the options to the user for setup.

ab7pa:
Closing this PR.

PR #357, opened 2022-05-06
enhancement: advanced config categories

dman776:
provide categories structure to advanced config page


ab7pa (approved):
Code works on my test nodes.  THX

PR #360, opened 2022-05-11
Link Quality Management

aanon4:
This is how I imagine the LQM would look if it were deployed as part of the standard firmware.

It's enabled by default, but if you disable it (advance settings) the system reverts back to the older distance UI and mechanism.


mathisono:
I feel the current Status page need a little more to help identify the condition of a node's rf link... I've fielded a lot of questions regarding what "LQ" and "NLQ" are!  The TxMbps value isn't really an accurate account of the links throughput!  With Tim's new values "Quality" and "TX Estimate" is much more informative and the information should be considered for placement on the status page. being displayed along side TxMbps would better inform users of links true condition. Is "TxMbps" really descriptive of its function?   I wonder if the Tim's method could be applied to periodically test tunnels and DTD "I have DTD vlan 2 being passed on a AC gen2 and AirFiber" so a periodic test would depict the status DTD and tunnel connections better than 100% LQ and 100% NLQ  for users.  Tim great work! This sets up a mechanism to asses the working quality of a link. What OLSRD doesn't do! Very usefull for BATMAN.

ae6xe:
@mathisono   The LQ/NLQ/txmbps  values have  help info, for sure in the on-node help and I think in the read-the-docs as well.   Is there anything we could do to improve the current docs, to refer people to when asked?     As I'm sure you know, TxMbps is a raw protocol rate when bits are transmitted, and the node is half-duplex also sharing the freq with potentially many other devices also looking to transmit data.   This is a layer 2 raw protocol rate, very different than a layer 3 throughput value  -- comparing apples and oranges.

Orv:
As it is right now, anyone can get to an LQM page and modify parameters without credentials.   Seems like this should require root access for modification, otherwise any random person could potentially take a node off the air by maliciously modifying the parameters.

aanon4:
Just addressing @Orv (because I know the answer to this one!) the actual proposed integrated UI (which is in this PR) isn't the same as the experimental UI. The configurable parameters move to the setup page and advanced config - both password protected.

Orv:
Good to know - thanks.

aanon4:
They status page and config currently looks like this:

Screen Shot 2022-05-11 at 5 31 51 PM

Screen Shot 2022-05-11 at 5 32 14 PM

ab7pa:
On the **advancedconfig** settings, what would you think about renaming the lqm.low_snr to lqm.min_snr just for consistency?  The other similar settings are named min_distance and min_quality.

ab7pa:
On /www/cgi-bin/lqm for the header, location, description, & <HR>.....  If there is no node_description the other pages skip it, but on **lqm** there is a narrow gray bar displayed even when the node_description is missing.  Not sure if you might want to make this consistent with the other pages.

aanon4:
Fixed advanced config and the description.
I don't want to clutter the UI with different elements to adjust the same distance value, but I take the point about miles being US/UK centric. It might be possible to do the necessarily localization in the browser (the node has no idea) so I will take a look.

ab7pa (approved):
This looks good to me.  tested between four nodes locally and everything worked as advertised.  Many thanks, @aanon4 Tim.

aanon4:
Update screen shots:
Screen Shot 2022-05-16 at 11 37 09 AM
Screen Shot 2022-05-16 at 11 40 36 AM
Screen Shot 2022-05-16 at 11 43 49 AM

mathisono:
The "TX Estimated" & "Quality" would look good next too LQ/NLQ/txmbps on the main status page.  "txmbps" is documented but I feel its lacking a descriptive title like "tx Mode mbps" descriptive context for novices users! I'm used to CINR method found in U's stock fW.  Carrier noise detection is most useful, I dont know if the open driver reports back like the AC gen2 driver. How much could be useful ??

The manger gathers very informative information. A simple grade ( unstable -->stable, or poor<->good <->great) / throughput measurement/tx Mode mbps.  
Allowing the manger to test it's links based on, raw bit count, a periodic iperfs testing, ping latency, etc. A test that's not dependent on the wireless driver would be able to grade and drop tunnel links on a server and DTD links as well.

Im so happy that Tim is improving the metrics that the fw is using to describe a link. AREDN may need to back this up with a new wiki. to reinforce how importin it is to build for a  stable and high CINR.
https://help.ui.com/hc/en-us/articles/360013260554-airMAX-Improving-SNR-CINR
https://en.wikipedia.org/wiki/Signal-to-interference-plus-noise_ratio

reading leisure..... 
https://www.mdpi.com/2079-9292/11/8/1227/pdf 
 
73

Orv:
There is in fact a periodic link ping test in the current iteration of LQM; if the latency is too high, it pulls the link quality down incrementally.

aanon4:
So I don't disagree what the LQM information would be good with the current main status information, but when I looked at doing this it would be difficult to fit within the current UI there, both in terms of visual real estate, as well as how that UI behaves when nodes move from "current neighbors" to "previous". In the end I didn't think I could make it work, so punted this information to it's own, new page. I would like to revisit this, probably under the umbrella of the new UI where we might want to separate up the "mesh" info page from the "neighbor" info (maybe?).

I *think* with a newer OpenWRT code (on the roadmap) we might get some more useful information out of the driver - at least, I don't see it in the current /sys/... file system but there are various interesting features documented in later kernels. These include spectral graphs of the radio environment which I think we could do something interesting with. It would be nice to approach the level of information availability found in airMAX.

I know there is a desire to include more tools in AREDN and expose them in a friendly way via the new UI. There's a balance to be met between how much to automate this and let the HAM tweak. LQM does about the minimum it can get away with to do a "good enough" job with link quality. It doesn't run iperf automatically because of the bandwidth consumption.

I don't think of this as the end of their process so much as a starting point. The idea of being able to classify a link with simple "good, great, unstable, ..." is a good one, but I'd like to see how well this performs, and what needs tweaking, before taking more steps. I think this should work well, but it's already reached a level of complexity that should be tested more widely before doing more.

dman776 (approved):

k6ccc:
I like the new Neighbor Status page - but there is one problem.
The Neighbor node names are a link, but the link does not include ".local.mesh" resulting in not being found.
Unlike the Mesh Status page, where anything with a link includes ".local.mesh"
Otherwise, great work!

PR #370, opened 2022-05-23
LQM fixes 4

aanon4:
Tidy LQM status:
  Remove TX Estimate which was duplicating information on the mesh page and confusing folk.
  Sort by name to stop the display jumping around.


dman776 (approved):
Looks like a good change

PR #374, opened 2022-05-26
Really old sysinfo.json dont have link_info

aanon4:
Very old APIs don't return link_info, and the code failed to handle this.


ae6xe (approved):
looks good.

PR #375, opened 2022-05-27
LQM fixes 5

aanon4:
Fix empty initial lqm status.

Limit distance between DtD nodes which are considered at the same site.
Some network setups use non-ham radios to connect nodes over DtD links.
These should not be consider the same site, so we limit how far appart DtDed
nodes can be when optimizing.


ae6xe (approved):
Good enhancement.

PR #382, opened 2022-06-05
add hardware type to Status page

ab7pa:


ab7pa:
Will now look like this:

Screen Shot 2022-06-04 at 8 15 15 PM

dman776 (approved):

PR #383, opened 2022-06-05
report node hardware better on Status page 06/07/2022

ab7pa:


ab7pa:
Improves the way Ubiquiti node models are displayed:

Orv (approved):

dman776 (commented):
Is the value for this "hardware type" to exact same on the admin page??  If not, we need to change this label (because hardware_type has a specific meaning)

ab7pa:
It's a way for anyone on the network to see the type of node hardware on a non-protected page (so they do not need password access to see the type of node hardware).

dman776 (approved):

PR #384, opened 2022-06-06
restrict msg banner to nonprotected pages

ab7pa:
This PR separates the normal alert banner for unsupported hardware (which will appear on every page), so that the AREDN Alert Messages will only appear on the non-protected pages (Status and Mesh).


ab7pa:
Now the messages banner will only be displayed on the Status & Mesh pages, keeping the setup & admin pages uncluttered.

Orv (approved):

dman776 (approved):

PR #396, opened 2022-06-13
Fix display of unknown radio model

aanon4:


ab7pa (approved):
This fixes the case where the radio name is unknown.  Thanks very much, @aanon4 Tim!

ae6xe:
This adds a layer to check if /etc/radio.json has an entry for all AREDN supported devices.   This is a good thing, to clarify the message given.     But in this example, there is no radio.json entry for this specific ubiquiti device.   It appears to be a new device with a new ID that we've not seen before.    We'll need to review the support data file to make sure this is the right image (loco-xw) to use vs the rocket-xw image used by the other PBE 400/600 devices.   It's possible Ubiquiti put out another hardware rev and upgraded some chips -- and the rocket image is needed instead for everything to work properly.

Orv:
Did anyone add the boardid for for this model to radio.json?  I told the user it'd be in the nightly build.  If it's not gonna happen, I'll adjust his expectations :-)

aanon4:
This fix will only stop the page from failing. I don't know the ins and outs of doing the radio fix itself, but perhaps @ae6xe does?

ae6xe:
I'll submit before tonight's build...

PR #398, opened 2022-06-15
Lqm2

aanon4:
Towards the end of working on the original LQ, it became clear that various groups run DtDs and Tunnels over radio links. And while we don't have direct access to a lot of those radio metrics, we can at least measure the quality of them. So this extends LQM to monitor all links using whatever data we can.

### DIFF
--- a/files/etc/config.mesh/aredn
+++ b/files/etc/config.mesh/aredn
@@ -24,7 +24,7 @@ config tunnel
         option maxservers '10'
 
 config lqm
-        option enable '0'
+        option enable '1'
         option min_snr '15'
         option margin_snr '1'
         option min_distance '0'
--- a/files/etc/uci-defaults/80_aredn_lqm
+++ b/files/etc/uci-defaults/80_aredn_lqm
@@ -4,7 +4,7 @@ do
     if [ "$(/sbin/uci -c $c -q get aredn.@lqm[0].enable)" = "" ]; then
         /sbin/uci -c $c -m import aredn <<__EOF__
     config lqm
-        option enable '0'
+        option enable '1'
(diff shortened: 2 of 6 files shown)


aanon4:
This is one of the new behaviors. The DtD link is clearly not a local connection to the other node (being 33 miles away) so LQM no longer assumes it hardwired and reliable. If it was close you'd see the old behavior.

ab7pa:
Got it!  Thanks for the clarification.

ab7pa (approved):
Tested between seven nodes and works as designed.  Thanks, @aanon4 Tim.

ae6xe:
@ab7pa  wrote, "Should collocated DtD nodes have their RF status set to blocked - dtd instead of idle? I'm not clear on which would be correct. THX" 

Yes, I think so.   The dtdlink path is optimal and redundant RF path is not necessary.   No need for OLSR to show this duplicate (and higher cost) route path across the mesh network.   Although, if the dtdlink path fails, we would want the RF path to not be blocked.

aanon4:
My reasoning for long DtD not to be treated as equal to short DtD is the the long ones are also RF and so subject to the vagary that come with. In these cases I thought it best to leave it to OLSR to pick a route (assuming they're not blocked for other reasons).

ab7pa:
Your explanation above cleared it up for me, Tim.  In the environment you have for all your testing you've seen many types of links that I have not seen here.  And I think you're correct to distinguish between long vs. short DtD links.  Makes sense because some of them are over radio.

Orv (approved):

ae6xe (approved):
looks good.

PR #399, opened 2022-06-15
Keep 100mb as the only speed, but let the port autonegotiate with the switch

aanon4:
Force device speed to be 100Mbps, but allows it to actively negotiate with the switch. This avoid any extra switch configuration, and is invaluable when the switch is dumb (like the popular outdoor nanoswitches) and cannot be reconfigured.


ae6xe (approved):
looks good.

PR #403, opened 2022-06-17
Fix error when re-enabling wifi mesh with lqm on

aanon4:
When re-enabling Mesh Wifi with LQM already enabled, there are no parms for LQM set at that moment so don't check them.


ae6xe (approved):
looks good.

PR #406, opened 2022-06-21
Remove #NOPROP from /etc/ethers where it was being incorrectly parsed.

aanon4:
The AREDN code now cross references entries in /etc/ethers with those in
/etc/hosts to determine if an IP is #NOPROP. The dnsmasq code could not handle
the #NOPROP flag at the end of /etc/ether entries.


ae6xe (approved):
looks good.

Issue #415, opened 2022-06-26
STXsq Lite2 no neighbors with LQM enabled


AJ6GZ:
With LQM enabled, STXsq Lite2 does not stabilize its links with other nodes. During this time LQ/NLQ is 100/0% with all RF neighbors, occasionally rising but falling back to zero. I noticed these models do not work with a distance Auto (0) setting while LQM is OFF and it exhibits the same behavor as above. Works fine with LQM off and non-zero distance.  Tested several times and it did come up and stabilize in about 20 minutes once, other times not at all. Tried with two different SXTsq's and Nano's in the mix. 
I can put a spare on the SoCal network if someone wants to poke around.

MikroTik SXTsq Lite2 RBSXTsq2nD
Build 1449 (and others prior)
Ch -2/10Mhz
[supportdata-AJ6GZ-Portable1-SXT2-202206242008.zip](https://github.com/aredn/aredn/files/8987632/supportdata-AJ6GZ-Portable1-SXT2-202206242008.zip)

aanon4:
Could you post a screenshot of the Neighbor status page with LQM enabled (after the status's no longer say 'pending')? Thanks.

aanon4:
This node doesn't have its location set?

ab7pa:
My Mikrotik SXTsq Lite 2 nodes link well with all other manufacturer devices, at least on channel -2 (5 MHz width) which is what our net is running.  They have worked with LQM enabled or disabled -- both modes work fine, and auto-distance (zero) also works.

AJ6GZ:
Been up for an hour... location not yet set
![image2](https://user-images.githubusercontent.com/42795596/175837120-149d9ae0-3948-4e09-ad83-05bee8dc6a9c.png)

Location set, after 20min uptime. Still no good connection.
![image3](https://user-images.githubusercontent.com/42795596/175837854-01bf0943-8a81-4956-abd3-eb66359a9cd7.png)

(That one neighbor is expected to the blocked)

aanon4:
Given the SNRs I assume these other nodes as essentially in the same place? What does their neighbor info look like?

AJ6GZ:
Yes normally two nodes on-site, one down the street.  The 3rd one being tested is with a portable setup powered on.  However the other fixed SXT2 does the same thing.

Basically the opposite LQ/NLQ at any given time... happened to be 10% at this moment... as seen from the Nano:
[...]

PR #416, opened 2022-06-27
Allow LQM auto-distance to be overridden

aanon4:
Allow auto-distance to be overridden when LQM cannot determine the distance to other nodes

https://github.com/aredn/aredn/issues/415


ae6xe (approved):
Looks good.

PR #418, opened 2022-06-29
fixes for NAT Mode DHCP address range

ab7pa:
Prevent `setup:398: attempt to perform arithmetic on global 'dmz_dhcp_end' (a string value)` message.
Reported in this forum post: https://www.arednmesh.org/content/navigating-cgi-binsetup-gives-502-bad-gateway-status-code


aanon4:
This change bothers me more - why does this particular value need a "tonumber" but not the others? Feels like there's a root cause here that I don't understand.

ab7pa:
Closing this PR since it does not provide a fix for the symptoms that were reported in the Forum.

PR #422, opened 2022-06-30
allow zero length wan wifi client passphrase

### LINKED ISSUE #421: WAN WiFi Client should allow zero length passphrase
The WAN WiFi Client section should also allow a zero-length passphrase. Legacy code had this test: 
`if( "$wifi3_enable" eq "1" and (length( $wifi3_key ) < 8 or length($wifi3_key) > 64) and ! length($wifi3_key) == 0)`

So the new Lua code should allow a zero length passphrase too:
line 685: ` if #wifi3_key < 8 or #wifi3_key > 64 then`     
(the test should include #wifi3_key ~= 0 )

ab7pa:
Fixes #421

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -682,7 +682,7 @@ if parms.button_save then
         end
     end
     if wifi3_enable == "1" then
-        if #wifi3_key < 8 or #wifi3_key > 64 then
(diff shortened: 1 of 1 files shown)


ae6xe (approved):
good catch.  This accounts for an AP that does not have a password.

PR #423, opened 2022-07-01
Heat map spectral view

aanon4:
Some local analysis of the spectral data to reduce what needs to be sent over the wire (so page loads much quicker) which also allows a "heap map" showing how many transmissions occur at what power and at what frequencies.


ab7pa (approved):
Works perfectly!  Nicely done, @aanon4 Tim!
Screen Shot 2022-07-01 at 7 44 30 AM

PR #426, opened 2022-07-06
Fix tacticalname

### LINKED ISSUE #420: nodename nil value on Setup page
A [forum post](https://www.arednmesh.org/comment/19732#comment-19732) identified a "Bad Gateway" message on the **Setup** page that resulted from the _node_ value being nil.  This occurred using firmware ≥ 3.22.6.0.  Message text:
`Sat Jun 25 08:41:58 2022 daemon.err uhttpd[938]: /usr/bin/lua: /www/cgi-bin/setup:656: attempt to index global 'node' (a nil value)
`

Possibly include a case on line 656 that tests for the possibility of a nil value during nodename validation?  This would prevent the generic "Bad Gateway" message, but we also need to investigate what could cause the node variable to be nil within the "validate and save configuration" section of the code.

dman776:
fixes #420

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -642,7 +642,7 @@ if parms.button_save then
     end
 
     if nodetac:match("/") then
-        node, tactical = nodetac:match("^%s*([%w-]+)%s*/*%s*([%w-]*)%s*$")
+        node, tactical = nodetac:match("^%s*([%w-]+)%s*/%s*([%w-]*)%s*$")
         if tactical == "" then
             err("invalid node/tactical name")
         end
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -642,7 +642,7 @@ if parms.button_save then
     end
 
     if nodetac:match("/") then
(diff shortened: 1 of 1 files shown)


ae6xe (commented):
In LUA it appears that both "*" and "-" mean "zero or more repetitions".    To match a literal "-" within the host or tactical names, maybe this needs to be  "[%w\\-]".   but surely this is already working for hostname.

ae6xe (commented):
Shouldn't the change be to add the "+" -- needs at least 1 (probably more) characters to be a tactical name.    This could match and return an empty string as the hostname?  Maybe that's OK if there is another check later on to qualify the tactical name.   (using a '+' on hostname, to follow suit.)

ab7pa:
Screen Shot 2022-07-07 at 9 06 30 AM

ae6xe (approved):
looks good.

ab7pa (approved):
I did not reproduce the original issue, but I tested this on one of my nodes and it worked fine.

aanon4 (approved):

PR #427, opened 2022-07-06
rename supportdata extention from tgz to gz

dman776:


aanon4 (approved):
I assume this is to make attaching them to Issues easier?

dman776:
Yes

PR #429, opened 2022-07-12
Reduce colors and improve scan information

aanon4:
Reduce the graph colors and add a key to improve the the information being conveyed by the spectral scan. Now we display 3 bands - noise, 95% of traffic, and 5% of traffic. The goal is to convey both where the spectrum is in use, and how much it is in use. Hopefully this does a better job of that.

reduce-colors

### DIFF
--- a/files/www/cgi-bin/scan
+++ b/files/www/cgi-bin/scan
@@ -261,8 +261,15 @@ local cheight = 400;
 html.print([[
     <center style="font-family: Arial">
     <div style="padding:8px">Spectral View</div>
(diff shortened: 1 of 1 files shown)


ab7pa (approved):
Installed and tested on two RF connected nodes -- looks very readable.  Thanks @aanon4 Tim.

aanon4:
I've seen some reports of previous versions not working on 3GHz nodes, but I did test this on my M3 NanoBridge and it worked there.

Orv (approved):

PR #432, opened 2022-07-14
Fix spectral info arrays

aanon4:
If buckets only have a few samples, it could end up trying to read the 0'th one, not the 1'th. Lua has 1-indexed arrays.


ab7pa (approved):
This change seems to adjust the scale so traffic appears in the correct buckets.  Thanks, @aanon4 Tim.

PR #435, opened 2022-07-17
Spread 56 samples over 64 buckets (per the spec) - do not merge

aanon4:
Putting this up so people can compare spreading 56 samples over 64 buckets vs the current version which treats the samples as evenly spaced. In both cases we're inventing data we don't have - either treating the 8 missing samples as 0's, or spreading out our 56 samples.

sixtyfour

### DIFF
--- a/files/www/cgi-bin/scan
+++ b/files/www/cgi-bin/scan
@@ -310,7 +310,7 @@ end
 local bw = 10
 
 local fbuckets = {}
-for freq = 1, (end_freq - start_freq + bw) * 56 / bw
+for freq = 1, (end_freq - start_freq + bw) * 64 / bw
 do
     fbuckets[math.floor(freq)] = {}
 end
@@ -347,7 +347,10 @@ do
                         if sig > max_sig then
                             max_sig = sig
                         end
-                        local bidx = math.floor((freq - start_freq) / bw * 56 + dptr - 28)
+                        local bidx = math.floor((freq - start_freq) / bw * 64) + dptr - 28
+                        if dptr > 28 then
+                            bidx = bidx + 1
+                        end
                         local bucket = fbuckets[bidx]
                         if bucket then
                             bucket[#bucket + 1] = math.floor((sig - min_sig) * 4)
@@ -388,7 +391,7 @@ html.print([[
     for (let i = 0; i < p.length; i++) {
         let v = p[i];
         if (v === null) {
-            xcursor += ]] .. (xscale * bw / 56) .. [[;
+            xcursor += ]] .. (xscale * bw / 64) .. [[;
             ycursor = 0;
             ccursor = 0;
         }
@@ -396,7 +399,7 @@ html.print([[
             v *= yscale;
             ycursor += v;
             ctx.fillStyle = bcolors[ccursor];
-            // ctx.fillRect(xcursor, ]] .. cheight .. [[ + ycursor, ]] .. (xscale * bw / 56) .. [[, -v);
+            // ctx.fillRect(xcursor, ]] .. cheight .. [[ + ycursor, ]] .. (xscale * bw / 64) .. [[, -v);
             ctx.fillRect(xcursor, ]] .. cheight .. [[ + ycursor, 1, -v);
             ccursor++;
         }


ae6xe:
In 10 and 20MHz channel widths there are samples for every slot of the 64 in each channel, except on the band edges.    Thus, no data is made up, rather just not the same count of data in every freq bucket.  The 5MHz channel width is where there should be no samples for 7 freq buckets between channels.  Is it possible to show no information or no color (background color) in the freq buckets for which there is no data?  

before it was like we had 40/20 eye sight and the picture was blurry.   Yes we can see there is a mountain in the picture, but this change makes it a 20/20 view of the picture.

aanon4:
If I were to set my device to a 5Mhz channel size and rerun this test, the resulting scan would have gaps. That isn't what's actually going on in my environment though, because here there are devices transmitting with 10Mhz and 20MHz bands, and there is absolutely transmissions in those "blank" spaces. Showing blanks would convey a false impression of what's actually happening. Similarly with wider band sampling we get few samples in some spaces than others which ... again ... isn't actually what's happening.

In all cases, because the transmitters and scanners are not necessarily operating on the same bandwidths or the same central channels, and we only sample 56 of the 64 subbands per scan, we know that we are missing 8 bands of data which may, or may not, have actual information in it. As such, any representation of the scan data is a compromise. I just think the simplest compromise does a good job and adding the gaps in doesn't improve the representation of what's going on.

I but this branch up to encourage people to try this out on your own radios with different settings, see how it looks, and compare it to what's currently in the nightly.

ae6xe:
This isn’t the case for 10 and 20MHz channel width settings.    While 8 of 64 data points are not provided in the scan at a given channel, the scan of the adjacent channel does have this missing data, because the channel freq space overlaps.   We are not making up data across the band and correctly showing in the chart the exact freq where all the samples were measured from.  

 In a 5MHz channel width setting, the radio does not capture samples or data at a lot of freq space.  It’s a given there’s something that could be measured there, but the radio is not capable at this setting.   Our choice is to make up data OR show no data.  It is a given, there is noise or even a signal that will be at these freq locations.      My vote is to show no data, not even noise.     but I put out to the group of what to do.  We had the same question come up when the snrlog samples came back with an error or no data sample.

aanon4:
On the 5MHz point, if I see bands of data with blanks between them, I'm going to think something is going on .. and it isn't because those blanks are an artifact of the scan. It's misleading to no purpose.

I encourage people to try this code and the nightly code and see which they prefer in practice.

ae6xe:
Summary of options for 5MHz channel width setting, and what to do because the radio is not measuring 12.5% of the band;   1) show as noise floor, 0 snr. ; 2) streach adjacent measured frequencies into the space (prior technique);  3) extrapolate from adjacent measures through gaps (effort and complexity to do); 4). Don’t show any data for freq space there are no samples measured (holes in the spectrum chart)

ab7pa:
Here is a screenshot from testing the code in this PR, using a 2.4ghz node on channel -1 at 10mhz width.  The graph in this scan seems to track fairly closely with the wifi traffic at my QTH -- most traffic is clustered around channels 1, 6, and 11.  At times the different colors confused me, so I used a single color for the graph.

Screen Shot 2022-07-17 at 5 29 39 PM

aanon4:
This is why I dislike this approach - you get gaps in the graph which aren't really there (I think home wifi has 20 or 40 Mhz wide channels .. mine does anyway). Makes is look like something is going on when it isn't.

aanon4:
I put this up here purely so people could try it out - it is not design to be merged into the build.

ae6xe:
There should only be 1 gap centered on  ch 0, since this is not a channel selection option to obtain spectral measurements and channel 1 and -1 scans do not return samples at this location.

there appears to be a gap centered on odd channel numbers.   There should not be gaps here.   Something is amiss — should not appear as a different color or no color.  There are data samples on the odd channel numbers, the same as even channel centers.  The center of all channels would have less data samples, than elsewhere in the band, but there are samples to show.

aanon4:
Can you post what you see when you try it?

ae6xe:
“ One of the things that happens with any of the approaches is that the scan samples can vary widely from scan to scan. This can result in somewhat inconsistent results across a series of scans.”

the way to deal with this issue is to increase the spectral_count to obtain sufficient samples of data over enough time that the signal picture is more complete.    Try adjusting this value, is there a count that the inconsistencies start to go away, and the time to scan is still tolerable?

ae6xe:
It would be better to think in terms of changing how to display the data and move this PR forward to resolve the issue.   For 20MHz chart, where there are no gaps, this PR shows the results exactly on the frequencies measured from, it is correct at 20/20 vision.  The previous code, the chart is blurred and it’s like we’re looking at the same data with 40/20 vision — the samples are stretched around to other freq locations.  We just need to figure out how to display 1 gap in the data samples on 10MHz chart, and several gaps on the 5MHz chart.   We have more options with this PR to decide how to cover the gaps, and it corrects the vision to 20/20.

ae6xe:
> There should only be 1 gap centered on ch 0, since this is not a channel selection option to obtain spectral measurements and channel 1 and -1 scans do not return samples at this location.
> 
> there appears to be a gap centered on odd channel numbers. There should not be gaps here. Something is amiss — should not appear as a different color or no color. There are data samples on the odd channel numbers, the same as even channel centers. The center of all channels would have less data samples, than elsewhere in the band, but there are samples to show.

This is ref to the 10MHz chart.  I am forgetting about the location of 1 sub-carrier centered on each channel, so my prior comment as an error.   No data is encoded at this sub-carrier location, thus there's no spectral scan sample measure given at this location.  This PR is accounting for this issue, and as @aanon4 noted, we would see a 156.25kHz wide space centered at each channel #.   There is no overlap with the scan of other channels.  The spectral scan doesn't tell us what measure of power exists in this space.    However, there may still be a coding error with this PR, because we don't see the same line centered on both even and odd channel #s?  Maybe just a display issue, in the resolution of screen, etc.?

This PR shows the 20MHz channel width spectral scan the same as other 3rd party implementations and accurate to the frequencies as measured, with no gaps in data.

The gaps in data returned from a 10MHz channel width setting spectral scan is 1.09375MHz (space of 7 sub-carriers 10MHz/64*7) of space around ch '0' (on 2GHz band and 900Mhz band) and 156.25kHz (10MHz/64) space on the center of each channel (for all bands).  This PR, for a 10MHz channel width setting, accurately displays the samples at their measured freq locations with gaps where no data exists to display, with exception noted above.

The gaps in data returned from a 5MHz channel width spectral scan is 78.125kHz (5MHz/64) at the center freq of each channel and 546.875kHz space between each channel (5MHz/64*7).  Same 2GHz and 900MHz issue at ch '0' gap.

aanon4:
Closing this. It's here so people can try the code if they want to.

PR #440, opened 2022-07-19
Ping tool (to complement the iperf3 tool)

aanon4:
This is a simple ping tool to complement the existing iperf3 tool. It allows someone to initiate a ping test from any node to any other node. The test is limited to 10 seconds.

The goal is to enable improved network monitoring by allowing specific links in the network to be tested. Right now, although nodes can be tested, the path will always be from the initiator to the target, which is not necessarily what you want.

The use mirrors the "api" we use for the iperf tool.


ab7pa (approved):
Works well on my test node.

Screen Shot 2022-07-19 at 4 32 57 PM

dman776:
What do you think about adding a getPing function to aredn.utils and exposing in the api.   This is the same approach as the traceroute takes.   http://localnode.local.mesh:8080/cgi-bin/api?traceroute=TARGETNODENAME
it could be: http://192.168.0.47:8080/cgi-bin/api?ping=TARGETNODENAME

dman776 (changes requested):
see comment about adding to api

aanon4:
I had no idea that was even there!
Could do - was just using iperf as a basis for this, but could use traceroute instead - not fussed either way

ab7pa:
I'll wait to wrap up the documentation for this feature until we decide whether to accomplish it through cgi-bin/ping or cgi-bin/api -- either way it looks like a good addition.  I suppose that cgi-bin/iperf could eventually be converted to the api approach at some point.  Thanks, @aanon4 Tim!

dman776:
May as well move them to api for future.   We can always have a "html" readable page in the interim that essentially calls the api and formats for html if you think it's necessary.   That is my strong preference.

ab7pa:
One issue with using cgi-bin/api is that the results are displayed in json rather than in plain text in the browser.  This PR still seems valuable (as well as cgi-bin/iperf) for displaying the results in simple text.

dman776:
See previous comment

Thanks,
Darryl

dman776:
also, is the ttl value needed?   It's always going to be the same 64, right?

aanon4:
Bump.
TTL - I decided to include all the information I was given, because maybe it would be useful.

PR #445, opened 2022-07-21
Fix random iperf failures

aanon4:
Sometimes, when an iperf3 server is started for a test, it will bind to an address which isn't the main mesh address. For tcp this isn't a problem as the test is connection based, but for UDP this can cause issues by sending UDP from an unexpected address. By always binding the server to the primary node address we avoid this problem.

I've been chasing this for a while as it's been making automated UDP based testing unreliable.


ae6xe (approved):
looks good.

Issue #446, opened 2022-07-21
New installation on AR300M16 defaults to Channel -2 Bandwidth 20 MHz


pmilazzo:
When installing AREDN on a new AR300M16 (2.4 GHz), the node initially defaults to Channel -2, which is correct, but with a bandwidth of 20 MHz, which is not. I observed this behavior with both 3.22.6.0 and nightly 1510-ba55aed.

aanon4:
I think it's the old perl thinks 10 == "10" while in lua 10 ~= "10" problem.

aanon4:
Booted up a brand new AR300M and it came up with bw of 10.

PR #448, opened 2022-07-21
wifi_chanbw should always be a string

aanon4:
https://github.com/aredn/aredn/issues/446


ab7pa (approved):
Looks good on my test nodes.  Thanks @aanon4 Tim.

ae6xe (approved):
looks good.

Issue #397, opened 2022-06-14
MAC address duplicate for GL-AR300M


tpreston51:
I was trying to move me Tunnel Server and Tunnel Client to separate GL-AR300M devices to separate them from my active RF routers.
I found that the two delvices had the same MAC address on the br-lan and eth1 lan configs therefore causing problems with my switch.
Software 3.22.6.0 loaded but it also did not work on 3.22.1.0

aanon4:
Could you post the support tool data for the node?

tpreston51:
[Supportdata.zip](https://github.com/aredn/aredn/files/8904737/Supportdata.zip)

pmilazzo:
I can confirm this behavior on a pair of AR300M16-Ext nodes. When booting the manufacturer's latest firmware, I get:

pmilazzo:
### Workaround:
(fiddly, but it works)

1. Use the `ip link` command to determine the MAC address of eth0.
2. Add 1 to the hexadecimal value of the last byte to derive the correct MAC address for eth1. For example, if the address ends in ...:a2, use ...:a3
1. Edit `/etc/aredn_include/ethmacfixup` to insert a line of the form:
`            option macaddr "xxx"`
    where one or more spaces precede `option`, and _xxx_ is the MAC address derived above, surrounded by literal quotation  marks.
1. Edit `/etc/config.mesh/_setup` and change the value of `dtdlink_ip` to 10._x_._y_._z_, where _x_, `y`, and `z` are the _decimal_ equivalents of the last three _hexadecimal_ pairs in the MAC address derived above. For example, if the derived address ends in ...:a3 , then z=163.
3. Go to the _Basic Setup_ tab in the UI and press **Save Changes** and then **Reboot**.

aanon4:
So there's a file /etc/init.d/local which runs on each boot, but specifically on first boot it creates the /etc/aredn_include/ethmacfixup file based on a few rules.  Is this file empty on this hardware? Unfortunately it's not included in the supportdata. Maybe we need to add some new rules here for this device.

aanon4:
I just bought one of these on Amazon so hopefully can get to a solution this weekend.

pmilazzo:
That file exists, but it doesn't handle this case because the eth1 MAC address is valid and differs by more than 10 from the WiFi MAC address.

As an experiment, I changed the test on line 24:

    if [ $macdelta -le 10 -o $lanmac = "00:03:7F:11:23:C6" ]

[...]

PR #451, opened 2022-07-24
Deal with lan on AR300M always having the same mac address

aanon4:
https://github.com/aredn/aredn/issues/397

Credit to @pmilazzo for all the leg work on this.


ae6xe (approved):
looks good.

PR #454, opened 2022-07-25
Add WAN VLAN support to the Mikrotik hAP

aanon4:
The hAP is the first device to support which also contains a switch chip. In this case it's easy to do as the WAN port is separate from the switched port.

### DIFF
--- a/files/www/cgi-bin/advancedconfig
+++ b/files/www/cgi-bin/advancedconfig
@@ -465,8 +465,9 @@ function supportsVLANChange()
     if not (stat and stat.size > 0) then
         return true
     end
-    -- We also support VLAN changing on hAPs as WAN is on it's own ethernet port
-    if aredn.hardware.get_type() == "rb-952ui-5ac2nd" then
+    -- We also support VLAN changing on hAP, A750 and AR150 as WAN is on it's own ethernet port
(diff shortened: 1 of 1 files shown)


ae6xe (commented):
other devices in the same camp.  include in this PR?   ar750, ar150, ar300m16, airRouter.   However, is this PR adding value?   -- the WAN on these devices is untagged, and a manage switch could add the tag-ingress to be whatever internal switch vlan is desired (and remove the tag-egress).

aanon4:
Well I can't pretend this adds value for many users, any more than the general VLAN on WAN feature adds values for many users, but it's useful for some, and it brings this feature to more devices. From a network OCD perspective, having an untagged VLAN between switches is not good, so this tidies that up. I'm also looking at a scenario in SF where we're building a Backbone using Part 15 radio, and being able to tag the WAN from an hAP would be tidier than using an external switch to add it (the Part 15 radios provide a VLAN "WAN" in this scenario). 

We already support the VLAN on WAN on the AR150 and AR300M16. AirRouters have been out of stock when I've tried to buy them. I will add in the AR750 as it uses its switch in the same way as the hAP (just validated this).

ae6xe (approved):
looks good.

PR #457, opened 2022-07-26
Improvement to service checking

aanon4:
1. Add an 'arping' check for an address is a normal ping fails. This uses arp to find an address on the network, and will find devices which are live but have pings blocked. Only works for local lan devices.
2. Feedback from users - After an upgrade, services are left registered even though the packages are not yet reinstalled. We hide these service entries and restore them when the package is installed again - but there was a delay because we only run the check once per hour. It wasn't obvious what was happening, so now we immediately run the check after a package install.

### DIFF
--- a/files/usr/local/bin/olsrd-config
+++ b/files/usr/local/bin/olsrd-config
@@ -152,6 +152,7 @@ end
 
 -- validation
 if validate then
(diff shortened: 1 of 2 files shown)


ae6xe (approved):
Looks good.

PR #458, opened 2022-07-27
Add a visual indicator when a service is not being published.

aanon4:
Based on user feedback, and @ae6xe earlier comments, added some visual feedback when a service is not being published. Some hover text says what's going on, but there probably needs to be some help to explain why this might be happening.

Just reusing a 'dot' asset we already had.

### DIFF
--- a/files/www/cgi-bin/ports
+++ b/files/www/cgi-bin/ports
@@ -1207,7 +1207,7 @@ function print_services()
         else
             html.print("serv" .. val .. "_del value=Del title='Remove this service'>")
         end
-        local svc = string.format([[PlParam "service" "%s://%s:%s/%s|tcp|%s"]], _proto, _host, (_link == "1" and _port or "0"), _suffix, _name)
+        local svc = string.format([[PlParam "service" "%s://%s:%s/%s|tcp|%s"]], (_link == "1" and _proto or "http"), _host, (_link == "1" and _port or "0"), _suffix, _name)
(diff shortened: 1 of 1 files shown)


ab7pa:
After node upgrade/reboot the services list shows the grey dot to indicate that a service is not being propagated across the mesh.  The last entry for _MeshChat_ is valid because the _MeshChat_ package has not yet been installed after upgrade.  However the two dots for "placeholder" entries are also not being propagated when they actually should be.  Both of the hosts were pingable, so I'm not sure why they were marked for non-propagation.

Screen Shot 2022-07-27 at 6 31 22 AM

aanon4:
@ab7pa So it look like I just got the check for non-link services incorrect here? Unlike the code which runs hourly and works out if services *should* be advertised, this code is just checking to see if services *are* advertised (ie. in the olsrd.conf file). Looks like I may have the check for non-links wrong.

What do you think of the little grey dot? Open to other/better/different options here (UI is not my thing).

ab7pa:
Grey dot with hover text is fine.  I already have the documentation written and will submit the PR later today to include it.  Thanks @aanon4 Tim, for all your innovative solutions and super-useful enhancements to the firmware.

ab7pa (approved):
That second commit fixed the issue with unlinked "placeholder" services.  Looks good and works as designed on my test node.  THX @aanon4 Tim!

ae6xe (approved):
looks good.

Issue #463, opened 2022-07-31
[Feature Request] Toggle password visibility


aanon4:
In various password fields it would be useful to be able to see the text if desired.

PR #469, opened 2022-08-02
Password visible toggle

aanon4:
Add toggle for password visibility. Open to better design.
pwdviz


aanon4:
https://github.com/aredn/aredn/issues/463

aanon4:
@K6AH asks if someone could do a security review.

ab7pa (commented):
Works as intended on my test nodes.  Simply changes the input display type from _password_ to _text_ and back when toggled.  All other security features remain unchanged.

aanon4:
So this field never shows what the password is, only what has been typed in the box. And this page is protected so only someone with the password can view it.

K6AH:
Got it.  That's the answer I was looking for.  Thanks

dman776 (approved):
looks ok

nc8q:
Tim: Thanks.
I would love to have this on the Client Wi-Fi password entry box.
3s, Chuck

aanon4:
@nc8q This is already on that field too.

Issue #462, opened 2022-07-31
[Feature Request] With WAN wifi client, show when connection established


aanon4:
In a similar fashion to when tunnels show they're connected, it would be useful to see when the WAN wifi client has connected successfully.

PR #472, opened 2022-08-02
Provide status about state of wan wifi connection

aanon4:
wanwifistatus
One tick when connected to wifi, two ticks when wifi and internet.


aanon4:
https://github.com/aredn/aredn/issues/462

ab7pa:
Works great!  Icons and hover text change based on status of the connection to the local AP.  We could possibly use alternative wifi icons?  (just suggestion; no strong feeling either way)
notdone  Not Connected
done  Connected to Wifi without Internet
donedone   Connected to Wifi with Internet

dman776 (approved):
good addition

Issue #465, opened 2022-07-31
[Feature Request] Optional double-NAT WAN support


aanon4:
Sometimes the WAN is also a 10.X.X.X address which isn't compatible with our current setup. We could use an optional double-NAT to avoid this problem.

cTurtle98:
openwrt 22 supports multiple routing tables.

PR #473, opened 2022-08-04
Reduce footprint of WAN when attached to 10-net network [Investigation]

aanon4:
When the WAN interface is connected to a 10-net network, things will often break because the mesh is a 10.0.0.0/8 network. When this happens, this fix reduces the WAN network to a single gateway address, reducing the change of an address conflict and letting the WAN work correctly.

A WAN address conflict will then only happen if the WAN gateway IP or the WAN IP of this device are also on mesh IPs; but this is much more unlikely than an entire subnet, and similar to the chance of a new node address conflicting with one already present.

### DIFF
--- a/files/etc/udhcpc.user
+++ b/files/etc/udhcpc.user
@@ -1,0 +1,49 @@
+#!/bin/sh
+<<'LICENSE'
+  Part of AREDN -- Used for creating Amateur Radio Emergency Data Networks
+  Copyright (C) 2022 Tim Wilkinson
+   See Contributors file for additional contributors
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation version 3 of the License.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  Additional Terms:
+
+  Additional use restrictions exist on the AREDN(TM) trademark and logo.
+    See AREDNLicense.txt for more info.
+
+  Attributions to the AREDN Project must be retained in the source code.
+  If importing this code into a new or existing project attribution
+  to the AREDN project must be added to the source code.
+
+  You must not misrepresent the origin of the material contained within.
+
+  Modified versions must be modified to attribute to the original source
+  and be marked in reasonable ways as differentiate it from the original
+  version.
+
+LICENSE
+
+case "${INTERFACE}@$1@${router}" in
+    #
+    # If the WAN is a 10-net address, we remove the network route and leave just
+    # the default route to the specific gateway in place. This reduces the likelyhood
+    # of an IP class to a single address - no worse than your average new node on the
+    # network.
+    #
+    wan@bound@10\.*)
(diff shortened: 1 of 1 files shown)


aanon4:
Alternative fix for https://github.com/aredn/aredn/issues/465

ae6xe:
Since policy routing tables are in use, My understanding is that olsr has routes for individual IP addresses and LAN subnets,  that first takes precedence.    Then if the IP address is not on the mesh, it would finally get to the 10.0.0.0/8 default route and send it out the WAN.  I need to review this code a little more, but on a quick scan, I'm not sure that it changes any behavior.

aanon4:
As is, it work in some scenarios, but not enough. I'm current testing removing the 10.0.0.0/8 routes on the DtD and Mesh interfaces. I'll put that code up so you can also test it and let me know what works or you and what doesn't.

aanon4:
This is my current test setup, the WiFi address appears to clash with the WAN address, but with these changes it all works as you'd expect.
Screen Shot 2022-08-03 at 10 58 25 PM

aanon4:
I stuck [investigation] on the just in case I screw up and make it non-draft again.

aanon4:
Coming back to this after sleeping on it, and it wasn't clear to me why I needed to remove those 10.0.0.0/8 routes at all - so I put them back and it still works. Ignorning the OLSR routes, I have the following routing information:
```
default via 10.20.87.1 dev eth0.200  src 10.20.87.20 
10.0.0.0/8 dev eth1.2 scope link  src 10.21.87.216 
10.0.0.0/8 dev wlan1 scope link  src 10.20.87.216 
10.20.87.0/24 dev eth0.200 scope link  src 10.20.87.20 
10.162.190.192/29 dev br-lan scope link  src 10.162.190.193 
```
And the WAN works correctly as you'd expect it to looking at this. So now I'm not sure how to create a scenario which fails.

ae6xe:
I had thought that the only failure situation had been if there was a duplicate IP address between the AREDN network and on the other side of the WAN (with a 10.x.x.x allocation).   If this is the case, then the failure is only with the device with the duplicate IP addresses (or a single mesh LAN subnet with devices), could not talk to the other side of the fence -- routing won't route to the other side of the fence, limited to the duplication devices or single LAN subnet.    The only resolution is to remove the duplicate IP addresses.    Otherwise, I'm not aware of a failure condition or change we need to the routing. 

The other failure that happens, when the WAN side has 10.x.x.x allocation, is when one uses a VPN (on a LAN device of the gateway node, so no RF path).   When I connect to a corporate network that also uses a 10.x.x.x (which my company does).    My laptop's routing ends up sending all the 10.x.x.x traffic to the mesh node, and not through the VPN tunnel.  Possibly, there are other use cases like this?   When the WAN side does not use a 10.x.x.x address, the VPN works fine.  Scratched my head on this one for a while...

aanon4:
The duplicate IP problem I can see - but it's statistically quite unlikely and yet many people tell me the WAN 10.X.X.X is a problem.

I'm closing this for now. I thought there might be some obvious problem and a fix but I don't see it and need to get some actual real-world info from someone.

ae6xe:
Probably a perceived problem, not a real problem.  An assumption of what is more common, thinking AREDN and OLSR have routes based on the full 10.0.0.0/8 range.   This is not the case.   

For the greater audience reading this, OLSR only propagates across the AREDN network routes based on IP by IP address of node's interfaces, and a Mesh node's LAN subnet (of 4, 8, 16, or 32 IP addresses).   All other 10.x.x.x addresses route out an advertised gateway or local WAN if one exists.   The particular path across the AREDN network follows the order of the route policy tables and depends on what interface and/or mesh node the packet is coming from and going to.

Issue #475, opened 2022-08-09
We dont actually prevent traffic between LAN and WAN when "Prevent LAN devices from accessing WAN" is enabled


aanon4:
See https://github.com/aredn/aredn/issues/474 for more details, but it seems we don't actually prevent traffic between the LAN and the WAN when you check the "Prevent LAN devices from accessing WAN" box. All that does is decide whether to hand a default route to the LAN client. The LAN client can set its own default route and happily access the WAN. We should add a firewall rule to prevent this.

Problem validated on version 3.22.6.0 using an AR750 connected to the WAN with a Mac as a LAN client and a hand-set default route to the AR750.

WZ0C:
This is related to issue #26.

Issue #474, opened 2022-08-08
LAN Client not getting router from DHCP when "prevent WAN access" is enabled


WZ0C:
The client is a Reolink 823A camera.  It successfully gets a DHCP lease and is accessible and usable on the local LAN segment.  However, it is not accessible to clients on other mesh nodes because it does not have a route out of the local segment.  If a remote client tries to access the camera, the TCP SYN reaches the camera; however, no SYN|ACK is returned (no route on which to send it), and the remote client can not use the camera.  If a local client tries to access the camera, the handshake completes normally, and the camera can be accessed.

The DHCP responses to the client do not have Gateway addresses in the main headers, and when "prevent WAN access" is enabled, the option 3 router is not sent to the client.  Additionally, the option 121 and 249 routes are never sent to this client.  (And if the options 121 and 249 are force sent to the client, the client doesn't appear to use them.)

If "prevent WAN access" is disabled, then clients on other mesh nodes are able to access the camera without issue.

"Prevent access" not set:  (images 1, 2, 3)
![image](https://user-images.githubusercontent.com/110853434/183499934-723ce531-a7c7-4a38-bf8c-a52449445897.png)
![image](https://user-images.githubusercontent.com/110853434/183500842-19e6b590-c078-4828-b4ae-6d78e76c3a58.png)
![image](https://user-images.githubusercontent.com/110853434/183501204-e1b462f8-f27a-41c3-9656-56fac9d1dbbf.png)

"Prevent access" set: (images 4, 5, 6)
![image](https://user-images.githubusercontent.com/110853434/183501325-13288f42-e491-49f2-a302-be8d4e278083.png)
![image](https://user-images.githubusercontent.com/110853434/183501836-8d12b2e1-9944-44cd-ac2f-bcad6c53e494.png)
![image](https://user-images.githubusercontent.com/110853434/183502102-a0bc2238-d68a-4711-8610-cd59d3d36679.png)

WZ0C:
After further investigation, it appears there may be a design assumption that a lan client could be multi-homed and/or have a default route on another router.  And so the option 3 router would only be appropriate to send to the client if it was accessing the WAN through the node.  And so DHCP option 121 is used to send CIDR routes for 10/8 and 172.16/12 so the LAN clients can access the mesh.

[...]

Issue #42, opened 2018-10-23
Option to disable forwarding traffic destined to the internet


Orv:
From the recent SoCal mesh network meeting -
Formalize and make persistent via the UI, the firewall configuration to enable/disable Internet traffic through the node.

current command line commands:

ip rule del pref 20080
ip rule del pref 20080
ip rule del pref 30099
ip rule del pref 30299

Issue #26, opened 2019-03-27
Preventing LAN devices from accessing WAN doesn't work completely


n4tqu:
I upgraded to 3.19.3.0 today and decided to try the option for "Prevent LAN devices from accessing WAN" and noticed that this does not work as written. An existing LAN device will continue to access the WAN at least until the DHCP lease on the client renews and drops the default route. Also, anyone can add a static route on the LAN device and be capable of accessing the WAN.

I suggest an iptables rule would likely be more appropriate as a preventative measure.

Additionally, checking this box really should not require a restart of the mesh node. I suspect this is done due to limitations of the hardware however.

Steve - N4TQU

WZ0C:
It would be preferable for the UI option wording to be for a positive decision rather than a negative decision.  That is, "Enable Default Route" rather than "Disable Default Route".  A user having to turn something on (the checkbox) to turn off the feature (default route) can lead to confusion, frustration, and misconfiguration.   And similarly, "Enable LAN device access to WAN", rather than "Disable LAN device access to WAN".

PR #476, opened 2022-08-09
Fixes for "Prevent LAN devices from accessing WAN"

aanon4:
Part 1 - when "Prevent LAN devices from accessing WAN" is enabled, use a firewall rule to prevent traffic from LAN going to the WAN.

Part 2 - we need to change the dialog and the DHCP offerings to allow three cases:
a) LAN can access WAN
b) LAN cannot access WAN (LAN devices get a default route)
c) LAN cannot access WAN (LAN devices dont get a default route)

Option (b) is necessary for devices which dont support DHCP option 121 and 249 .. such as many IP cameras.
Option (c) is what you probably think you have now, you don't because traffic isn't blocked.

Just need some good UX/UI suggestions to finish this up.

Related to:
https://github.com/aredn/aredn/issues/475
https://github.com/aredn/aredn/issues/474
https://github.com/aredn/aredn/issues/26
https://github.com/aredn/aredn/issues/42

### DIFF
--- a/files/etc/config.mesh/_setup
+++ b/files/etc/config.mesh/_setup
@@ -33,7 +33,6 @@ dhcp_end = 25
 dhcp_limit = 20
 
 olsrd_bridge = 0
-olsrd_gw = 0
 
 wan_proto = dhcp
 wan_dns1 = 8.8.8.8
--- a/files/etc/config.mesh/_setup.default
+++ b/files/etc/config.mesh/_setup.default
@@ -27,14 +27,12 @@ lan_proto = static
 lan_ip = 172.27.0.1
 lan_mask = 255.255.255.0
 lan_dhcp = 1
-lan_dhcp_noroute = 0
 
 dhcp_start = 5
 dhcp_end = 25
 dhcp_limit = 20
 
 olsrd_bridge = 0
-olsrd_gw = 0
 
 wan_proto = dhcp
 wan_dns1 = 8.8.8.8
--- a/files/etc/hotplug.d/iface/11-meshrouting
+++ b/files/etc/hotplug.d/iface/11-meshrouting
@@ -42,7 +42,7 @@ fi
 
 if [ "$ACTION" = "ifup" ] ; then
 
-  is_olsrgw=$(cat /etc/config.mesh/_setup|grep -i olsrd_gw|cut -d ' ' -f 3)
+  is_olsrgw=$(/sbin/uci -q get aredn.@wan[0].olsrd_gw)
 
   # Set routes for wifi or device to device linking
   # unreachable rule is needed  to ensure packets do not traverse a rule later that considers routing to another network.
--- a/files/etc/local/mesh-firewall/10-lan-to-wan
+++ b/files/etc/local/mesh-firewall/10-lan-to-wan
@@ -33,14 +33,11 @@
 
 LICENSE
 
-noroute=$(grep "lan_dhcp_noroute" /etc/config.mesh/_setup | sed s/^lan_dhcp_noroute\ =\ //)
+lan_dhcp_route=$(/sbin/uci -q get aredn.@wan[0].lan_dhcp_route)
 wan=$(grep "wan_intf" /etc/config.mesh/_setup | sed s/^wan_intf\ =\ //)
 
-case "${noroute}" in
-0)
-    # LAN to WAN okay
-    ;;
-*)
+case "${lan_dhcp_route}" in
+  0)
     # LAN to WAN forwarding is disabled
     # Inserted in reverse order
     # Block traffic forwarding between LAN and local WAN (need this rule if WAN happens to be 10.X or 172.16.X)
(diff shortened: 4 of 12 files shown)


WZ0C:
Here's some thoughts for the UX/UI:

* Rename "Advanced WAN Access" to "Advanced Network Access"
* Rename "Allow others to use my WAN" to "Allow other nodes to use my WAN"
* Rename "Prevent LAN devices from accessing WAN" to "LAN devices need WAN access"
* Add "LAN devices need default route".  This item should be force-enabled if "LAN devices need WAN access" is enabled.

WZ0C:
Here's proposed updates for the help doc:

**LAN devices need WAN access:** Normally any devices connected to the node's LAN network will also have access to its WAN. Disabling this option will cause the node not to give LAN devices access to the WAN. This means that computers on this node's LAN will only use the 10.0.0.0/8 and 172.16.0.0/12 mesh network ranges. With this setting disabled you will not be able to access the Internet, even if your node has Internet available on its WAN port. This also applies to Internet available over the mesh. Disable this if your node needs to be connected to two networks at once, such as wired to the mesh as well as WiFi to a local served agency network.

**LAN devices need default route:**  The node provides routes via DHCP to give LAN devices access to to the appropriate networks.  All LAN devices by default receive routes for the mesh network ranges 10.0.0.0/8 and 172.16.0.0/12.  Some devices, such as certain IP cameras, may not support DHCP option 121 and will instead need a default route enabled in order to access the mesh networks.  A default route is also required for WAN access, and this is enabled automatically if "LAN devices need WAN access" is enabled.  If a LAN device is connected to two networks at once, such as wired to the mesh as well as WiFi to a local served agency network, care should be taken to understand how each LAN device handles having default routes to multiple networks.

ab7pa:
It will be important to get feedback from @ae6xe , @dman776 , @K6AH , and @wu2s Randy.
Personally it seems like the Advanced WAN features should be on the **Advanced Configuration** page where there is already a section for WAN Settings.

WZ0C:
It could also be simply "Network Access" rather than "Advanced Network Access".  The settings on Advanced Configuration seem a bit more technical and less aimed at the everyday user.

aanon4:
After discussions, migrate the LAN to WAN options to the advance options page. Really, the defaults are what most people want.

aanon4:
Here's a couple of screen shots of where the options used to be, and where they are now:
Screen Shot 2022-08-11 at 8 41 19 AM
Screen Shot 2022-08-11 at 8 40 55 AM

ab7pa (approved):
Verified that olsrd_gw ON **does** allow devices coming through Mesh RF to access the WAN/Internet.  Verified this works with lan_dhcp_route either ON or OFF.  Verified that lan_dhcp_route OFF **does** prevent local LAN devices from accessing the WAN/Internet.  Hopefully @WZ0C can test the third option on his camera that requires the default route.

WZ0C:
Received.  Building the image for test.

dman776 (dismissed):
Looks good to me.   @ae6xe what do you think?

WZ0C:
My camera works, but I don't think the feature is working correctly.  

In all four combinations of the lan_dhcp_route and lan_dhcp_defaultroute settings, DHCP option 3 is sent to the client.  I was unable to configure these two settings so that DHCP option 3 would not be sent.

The settings in /etc/config/aredn and /etc/config.mesh/aredn do correctly reflect the UI in all four cases.

aanon4:
Looks like I missed the callbacks to call "node-setup" on the new options (somehow I manage to remember for only one of them).

ab7pa:
@WZ0C , thanks so much for testing this with your camera, etc.  You should be able to retest by editing the _advancedconfig_ file to add the two lines indicated in the latest commit:  https://github.com/aredn/aredn/pull/476/commits/b94fc6665e3f6f7bc1fb4b103665bb5fa55bd042

WZ0C:
@ab7pa @aanon4 Rather than add the two lines, I tested with Tim's workaround of hitting Save config on the first of the three options.  

All four of my test cases of the two lan_dhcp settings being enabled/disabled passed.  If either setting was enabled, the DHCP response included Option 3.  If both settings were disabled, the DHCP response did not include Option 3.  

In my cases, none of the responses included Option 121 or Option 249, which were not asked for in the camera's DHCP request.  This is the same behavior as pre-fix, which is what I would expect in my testing.

Much appreciation to the group that implemented the fixes and helped me with builds.

ae6xe (commented):
pre-existing issue - LAN devices having a 172.16.0.0/12 route?    I don't think a LAN device would ever have awareness of, or need a route to a vtun 172.16.0.0/12 IP address.  All AREDN network addresses the LAN devices would be trying to reach are 10.x.x.x addresses.   Unless I am missing a use case here, can't we simplify and remove this 172.0.0.0/12 address in the DHCP route option?

ae6xe:
Looks good and needed.   AFAIK, This detail and logic hasn't been reviewed or really updated by any of us in 8+ years.   Well, other than trying to give a better name to describe what it was doing.

aanon4:
Removed 172.16.0.0/12.

ab7pa:
What are the implications for people using NAT mode for their LAN network?  Doesn't the IP address range for LAN NAT mode use a 172.x.x.x subnet?

aanon4:
While I hadn't considered this, I think we're fine. NAT mode means these addresses aren't exposed to the Mesh, so other devices on the Mesh wouldnt have these addresses to contact - so no need for a route to them and no need to hand LAN devices a specific route to them.

ab7pa:
Perfect!  Looks like we're all set, then.  THX

ae6xe (approved):
Looks good.

PR #477, opened 2022-08-09
reorder advconfig settings

ab7pa:
Put the _Advanced Configuration_ settings in a different order.  Since LQM is enabled by default, those settings could go first, followed by other system settings, with paths and alerts last.

[AB7PA-Hub Advanced Configuration.pdf](https://github.com/aredn/aredn/files/9293416/AB7PA-Hub.Advanced.Configuration.pdf)

### DIFF
--- a/files/www/cgi-bin/advancedconfig
+++ b/files/www/cgi-bin/advancedconfig
@@ -80,6 +80,155 @@ end
 
 local settings = {
     {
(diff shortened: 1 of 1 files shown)


aanon4:
If it makes any difference, LQM is only enabled by default on new installed, not upgrades.

dman776 (approved):

PR #481, opened 2022-08-15
A simple way to keep a user swconfig override across upgrades

aanon4:
Proposal for a simple way to allow a user to override the default swconfig and keep it across upgrades.
This seems sufficiently specialized that even an advance config is not the way to go; but as you can already
override swconfig if you want, this just provides a mechanism to keep those changes across upgrades.

### DIFF
--- a/files/etc/arednsysupgrade.conf
+++ b/files/etc/arednsysupgrade.conf
@@ -12,6 +12,7 @@
 /etc/config.mesh/vtun
 /etc/config.mesh/network_tun
 /etc/config.mesh/aredn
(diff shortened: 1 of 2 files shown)


aanon4:
As no one has said anything bad about this, taking it out of draft.

dman776 (approved):
I'm ok with this as a preservation mechanism.   Perhaps in the new UI, we can build a UI to create custom switch configs.

k1ky:
I wholly support that line of thinking.  Thank you for providing a method of preservation of settings across upgrades!

PR #483, opened 2022-08-22
Dont validate hostname in NAT mode.

aanon4:
Hostname is always valid for NAT but is never in the valid_hosts table.


ae6xe (approved):
looks good

Orv:
This appears to resolve the issue.

Issue #482, opened 2022-08-19
[Feature Request] Configure static routes that are persistent across reboots and firmware upgrades


dranch:
Creating this issue by request from KN6PLV to make sure it's not lost:

Request:  Is there a place that I can configure static routes on a HaP running AREDN that will be reloaded upon reboot as well across firmware updates?  

Tim response: There’s a file “/etc/local/mesh-firewall/59-custom-rules” which is just a shell script executed when the firewall is reloaded.
It’s preserved across upgrades - you could pop stuff in there but … kind of ..  bit messy.  There’s probably a more official openwrt way to do it involving files in /etc/config … but we mess with those quite extensively so off the top of my head I couldn’t say if custom changes are preserved or not

aanon4:
This appears to be the official way to add static routes - https://openwrt.org/docs/guide-user/network/routing/routes_configuration - so I just need to implement a mechanism so we can include extra routes when we regenerate our /etc/config/network file. I think we already have a way include files in our config files so probably not too hard.

PR #487, opened 2022-08-26
Optionally include static routes (and preserve them across upgrades)

aanon4:
Provide a more standard way to include extra static routes which will be preserved across reboots and upgrades.

https://github.com/aredn/aredn/issues/482


ae6xe:
Suggest to have an empty include file with comments of purpose and example entry, so users know what format to use (openwrt UCI).  Could even have a URL to this openwrt documentation.  

Any consideration (maybe future) UI ability to add the static route?     Otherwise, this is ready to merge in main?

dman776:
looks good to me.

aanon4:
The reason its draft is I'm away from home until Wednesday so can't run proper testing on this. If someone wants to do a build and make sure it works then I can make it ready.

ae6xe (approved):
looks good.

PR #488, opened 2022-08-26
Snapshot hostnames after updates so we have a consistent copy to display

aanon4:
When the olsr hosts are updated, the file gets truncated which can cause problems for the mesh display.

Here we use the olsrd update notification mechanism to first make a copy of the current hosts_olsr and then atomically move them to the new host_olsr.stable file. This guarantees that the rest of the system never sees a truncated set of hostnames.


ab7pa (approved):
Works as designed.  Good job, @aanon4 Tim.

dman776 (approved):
Tested successfully by Steve

PR #491, opened 2022-08-28
If redirect ends at an https link assume it is valid

aanon4:
We cannot check https links as we dont have ssl installed, so we just assume any redirect which ends in https://... is valid

### DIFF
--- a/files/usr/local/bin/olsrd-config
+++ b/files/usr/local/bin/olsrd-config
@@ -179,9 +179,15 @@ if validate then
                 valid_services[#valid_services + 1] = service
             elseif proto == "http" then
                 -- http so looks like a link. http check it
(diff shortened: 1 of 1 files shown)


ab7pa:
For some reason with this new olsrd-config code I see an error on /www/cgi-bin/mesh:
`line 320: bad argument #1 to 'lines' (/var/run/hosts_olsr: no such file or directory)`

aanon4:
Is olsrd not running? It makes that file.

ab7pa:
I assumed that my test node had gotten confused, so I went back to 3.22.8.0 and reapplied the new olsrd_config.  olsr daemon came up after reboot, and the /var/run/hosts_olsr file was written correctly.  All pages are working as designed.  Sorry for the firedrill.  Probably an error on mytest node

aanon4:
Thanks for rechecking. Rather a fire drill than a duff nightly.

ab7pa (approved):
Installs and works as designed.  Thanks, @aanon4 Tim.

aanon4:
We okay to move this into the nightly?

dman776 (approved):

PR #496, opened 2022-08-29
change support link to button

ab7pa:
Make Support Data download more prominent with a button rather than a link.  Everything else on the page is enabled using buttons, so this makes Support Data consistent with that.


dman776 (approved):
tests fine on my hapac lite

PR #498, opened 2022-08-30
clarify Wan Wifi Client label

ab7pa:
As requested in the forum, clarify the label for the WAN Wifi Client to indicate that it only uses the 2GHz radio.

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -1345,7 +1345,7 @@ if (phycount > 1 and (wifi_enable ~= "1" or wifi2_enable ~= "1")) or (phycount =
         end
     end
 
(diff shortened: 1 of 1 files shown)


ae6xe:
The wifi client was able to use the 2GHz and 5GHz radios on a dual radio device.   (One needs to turn off the mesh RF to have the 2GHz option show.)    I'm assuming this behavior came across from perl to lua conversion?     Could this be a request to clarify that mesh RF only uses the 2GHz radio (which is 802.11n)?   The 5GHz radio on the hap ac lite is 802.11ac.

ab7pa:
I'll double-check and will update as needed.  Thanks!

ab7pa:
@ae6xe Joe, you are correct and this change is not needed.  It would only affect two dual band devices, but this PR is inaccurate.  Thanks!

PR #499, opened 2022-09-01
add units to setup and advconf pages

ab7pa:
Add any missing **units** for fields on the _Setup_ and _Advanced Configuration_ pages.


files/www/cgi-bin/advancedconfig
```diff
@@ -175,7 +175,7 @@ local settings = {
         category = "WAN Settings",
         key = "aredn.wan.vlanid",
         type = "string",
-        desc = "Specify WAN VLAN #",
+        desc = "Specify WAN VLAN number; must be an integer in the range [10,4096]",
```
dman776, line 178: Is it 10-4096 or 1-4096?

ae6xe: It should be 1 to 4094.   The use case is to work around  switch constraints, and avoid to impose further constraints.   The fact that some switch vendors use vlan '1' for management purposes is a reason to use another ID.    

Note. the range is 0 to 4095 or 4096 total values.  0 and 4095 are reserved in the spec.   0 is used to give packet priority through a switch and considered untagged.   Don't recall what 4095 or 0xFFF is reserved for.

The constraint AREDN would have -- check what vlan dtdlink is in use, and prevent that from being a selection.  Or give an error if it is selected with the conflict.

ab7pa: Thanks, guys.  Change is made in latest commit.

ab7pa:
This PR has been requested in the forum too:
https://www.arednmesh.org/comment/20174#comment-20174

dman776 (approved):

PR #500, opened 2022-09-06
Fix bad match for NAT dhcp address.

aanon4:
Fix for this issue - https://www.arednmesh.org/content/unable-save-change-nat-mode-32280


ab7pa (approved):
Thanks for fixing this, @aanon4 Tim.  Works on my test nodes.  Good job.

dman776 (approved):
Tested by Steve successfully

PR #501, opened 2022-09-07
simplify AdvConfig display

ab7pa:
What would folks think of this attempt to simplify and clarify the Advanced Configuration display?

Screen Shot 2022-09-09 at 7 38 54 AM


aanon4:
I like it! I assume we want to make this more approachable?

mathisono:
So needed! The descriptive options had me confused. Now it better...

ab7pa:
From what I'd heard the cryptic key names were part of the confusion.  I'm wondering who might still want to see them -- if a developer, then they are available in the code, and if a normal user, then the _Save Setting_ message will still show them.

aanon4:
I spent a little time this evening mocking up adding the key back into the current design in various ways, and I'm not sure it helps anything having it there. The best I came up with was this:
189215346-3ca0fe33-2497-4be1-bba9-c6cc2c5177ee
It's there, but not emphasized.

ab7pa:
IMHO displaying the key might confuse normal users since they don't understand what it means or why it's displayed.

dman776:
I equate this to asking an advanced windows user to use the windows registry editor to find and locate a specific key to edit.  It needs to be shown.  Tim has a good compromise with the smaller, text for the key.  Remember the big red warning box at the top of this page?  If a user can't understand what a key name is, should they be on this page?

ab7pa:
How about something like this?  Changed screenshot in initial post to the current format with keys.

dman776:
That looks good

dman776 (approved):

PR #502, opened 2022-09-09
Stop node's LQM neighbors including itself

aanon4:
Nodes were occasionally including themselves as a neighbor. Let's not do that as it's confusing.


ab7pa (approved):
Installed and ran with the new lqm.lua -- everything works as designed.  I never saw this issue, but the new code should prevent it if it ever does occur.  THX very much, @aanon4 Tim.

dman776 (approved):

Issue #460, opened 2022-07-29
Node goes "deaf" after a few days/weeks.


AJ6GZ:
On a two month old SXTsq2, I am seeing the node go deaf, meaning the remote side sees it with 100% LQ / 0% NLQ and reports full strength signal/SNR, whilst the local side shows no connections. The chart is flat-lined.  A reboot or power cycle restores it.  The remote side comes back up without intervention.  I've see it 3 times now, twice it took about two weeks and this last one only 3 days since a power cycle.  24V power is stable.  The node doesn't crash or lose other functionality, seems just the receiver quits.  Support file taken during period of "deafness".
[supportdata-AJ6GZ-1-SXT2-202207290112.zip](https://github.com/aredn/aredn/files/9215317/supportdata-AJ6GZ-1-SXT2-202207290112.zip)

We also have a Rocket M5 XW in town that has been doing the exact same thing for over a year now, again about every 1-2 weeks (ai6bx-rm5-sector-llva.local.mesh).  That one is still in the broken state as of this writing.

ab7pa:
Ian, could you check the Channel Width setting on that node?  In one place it looks like you want to use a 10 MHz channel width on channel -2, but the _iw info_ in the support dump looks like it's using 20 MHz width.  If the channel widths are mismatched, then the two nodes will not link.  Let us know what you find.  THX
```
========== iw dev wlan0 info ==========
Interface wlan0
	ifindex 8
	wdev 0x2
	addr 2c:c8:1b:22:3d:7b
	ssid AREDN-10-v3
	type IBSS
	wiphy 0
	channel -2 (2397 MHz), width: 20 MHz, center1: 2397 MHz
	txpower 20.00 dBm
```

AJ6GZ:
It's always been 10Mhz on this one.  Interestingly, when I run that command on other nodes I also see 20Mhz everywhere even though they are all set to 5 or 10Mhz.

ab7pa:
Cool.  I assume your SSIDs also match on both ends.  Are you running Link Quality Manager (LQM) on either node?  Just making some guesses.

AJ6GZ:
Yeah all standard SSID, no other funky configs.  LQM is enabled on both but the Rocket pre-dates LQM when it started.  On that one it (correctly) reports "Currently no RF neighbor data available".

ae6xe:
[...]

PR #503, opened 2022-09-10
Simpler deaf node monitor

aanon4:
A proposed, simpler, deaf node monitor. It a node has zero stations associated with it, it will run a wifi scan every 5 minutes until that changes. The scan recovers a node. Running the scan when nothing is associated with the node seems harmless.


ae6xe (commented):
The most common use case, if not all instances, there is always 1 neighbor still communicating and one or more weaker neighbors are filtered out (with the noise floor raised).   If I read this PR correctly, it will only trigger when all neighbor's drop out?

The penalty to regularly trigger a passive scan on one channel should be very low.  The more frequently this is done, these increases the risk of a race condition and impact when people do mesh status and any other features that also do iw commands.   wifi scan or mesh status would be at risk to come back with no data.   I'm not sure if this is something to be concerned about or not.

aanon4:
> The most common use case, if not all instances, there is always 1 neighbor still communicating and one or more weaker neighbors are filtered out (with the noise floor raised). If I read this PR correctly, it will only trigger when all neighbor's drop out?

All the reports on https://github.com/aredn/aredn/issues/460 are for the node going completely deaf, not just losing contact with one instance. That's probably to be expected as I doubt anyone would notice if one of many nodes had disappeared when casually looking at their mesh status. 

> The penalty to regularly trigger a passive scan on one channel should be very low. The more frequently this is done, these increases the risk of a race condition and impact when people do mesh status and any other features that also do iw commands. wifi scan or mesh status would be at risk to come back with no data. I'm not sure if this is something to be concerned about or not.

There's obviously no way to know a user is about to use the 'iw' command in some form. The old code was no different here but perhaps the frequency of its use was different?

In the original issue https://github.com/aredn/aredn/issues/460 people observing all nodes disappearing and if this is happening between runs of the monitor, it wont defect this. Thoughts on how we can change that code to deal with this situation?

aanon4:
One option would be to *just* trigger the scan in this code when transitioning from hearing someone to hearing no one. So it would happen extremely infrequently?

Also, in relation to the current rssi monitor's behavior, we could easily add detecting this specific condition to that code?

aanon4:
Here's an alternate approach https://github.com/aredn/aredn/pull/505

PR #505, opened 2022-09-10
Add a wifi scan trigger for when the nodes detected becomes zero

aanon4:
Add an additional rescan trigger to the current system. Here if we detect the number of nodes going to zero (from non-zero), we trigger a scan. We only trigger the scan on this transition.

### DIFF
--- a/files/usr/local/bin/mgr/rssi_monitor.lua
+++ b/files/usr/local/bin/mgr/rssi_monitor.lua
@@ -58,6 +58,7 @@ if not file_exists(logfile) then
 end
 
 local multiple_ant = false
+local last_station_count = 0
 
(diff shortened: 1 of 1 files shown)


ae6xe:
looks good to me.   This PR would handle when ANI feature is breaking down altogether and the node has gone deaf to all neighbors. 
https://wiki.freebsd.org/dev/ath_hal%284%29/AutomaticNoiseImmunity

ae6xe:
Lowering the 3.5 threshold does not guarantee this condition would always be detected,  but should reduce occurrence.  This PR would always catch this scenario to recover the neighbor links.

ae6xe (approved):
looks good.

PR #506, opened 2022-09-11
Add more 3XX redirects + 401 authentication

aanon4:
Add missing 3XX redirects (rare, but apparently in use)
Add 401 to allow services which require basic/digest authentication (some webcams need this)


ab7pa (dismissed):
Tested on local node and works as designed.  Thanks, @aanon4 Tim.

dman776 (approved):

PR #510, opened 2022-09-21
add ntp update period to basic setup page

ab7pa:
Allow users to set the NTP update period from a dropdown on the Basic Setup UI

Screen Shot 2022-09-21 at 10 58 49 AM

### DIFF
--- a/files/etc/uci-defaults/82_aredn_ntp_period_migrate
+++ b/files/etc/uci-defaults/82_aredn_ntp_period_migrate
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# get current value of ntp_period if any
+if [ -e /etc/config.mesh/_setup ] ; then
+    period=$(grep "ntp_period" /etc/config.mesh/_setup | sed s/^ntp_period\ =\ //)
+fi
+
+# ensure /etc/config.mesh/aredn has ntp values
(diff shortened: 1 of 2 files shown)


aanon4:
Needs a way to persist this setting across upgrades.

ab7pa:
Upon **upgrade** the _update-clock_ script defaults to the _cron.daily_ location, so /www/cgi-bin/setup also defaults to _daily_.  If you later change the value to _hourly_, that setting will persist across reboots.  I'll work on after-upgrade persistence in a subsequent PR if that is agreeable.  If not, I completely understand and will keep this as Draft until I get that working.

aanon4:
I think it probably needs to be part of this.
Take a look at "files/etc/uci-defaults" which contains a set of scripts which are executed after each upgrade. You could probably add something here to do what you need.

ab7pa:
Have included persistence of the new setting across upgrades now.  Please let me know if there is feedback or suggested changes that would make this small feature more efficient, etc.  THX

ab7pa:
As an aside, if people want NTP updates to run continuously then they should [install Tim's meshntpd](https://github.com/kn6plv/meshntpd) package on their nodes.

Orv (approved):
Tested - works as advertised.

Orv (approved):

dman776 (approved):
tested by Orv.

Issue #54, opened 2018-07-15
Add ability for DtDlink capability riding over 3rd party RF links


ae6xe:
Submitted recently by Orv W6BI, and numerous discussions by AREDN team and other groups in the past.    Some of the scenarios for this feature:

1) usage of TDMA AirOS  P2P link (is expected to have higher thoughput than CSMA AREDN links unless channel is too noisy, if stuck in part15 space). 
2) existing part15 wireless links.

In such scenarios,  the 3rd party wireless networks would be configured to look to an AREDN node like a cat5 cable.  However, these RF links would have higher data loss and thus require different settings to behave as expected.   Existing DtDlinks over a cat5 settings, if used over an RF link, would drop out unexpectedly.   OLSR requires different settings.

The design discussed to date:
1) create a new vlan 3 for connecting  AREDN nodes over a 3rd party RF link
2) usage of 172.x.x.x similar to tunnel link address space to conserve 10.x.x.x address for mesh devices. (transparent to mesh users.)

Deliverables would ideally include:
1) updating switch config examples to include vlan 3
2) firmware updates
3) Doc on usage and 3rd party device requirements, e.g. the 3rd party network, could be used to connect many AREDN nodes  (many to many)

PR #514, opened 2022-10-03
Add support for extra network links to OLSR

aanon4:
More complex AREDN network utilize other networking technologies to create networks. At the moment this is not well supported. One solution has been to use the WAN network as the "other" network, and run AREDN tunnels across it. This works but isn't using tunnel for their intended purpose. 

Instead, this change allows for the creation of new network devices (most probably VLANs, given the hardware of current AREDN nodes) which are then included in the OLSR configuration and act just like any other AREDN network link. The links can be broadcast (e.g. a VLAN shared between many nodes) or point-to-point (a shared or non-shared VLAN, but with a specific AREDN peer).

The majority of AREDN users would never see this and it is intended for network admin building complex, large, networks. No UI will be provided. Notable, the approach is network agnostic, and expect more complex networking to be done elsewhere on external routers and switches.


aanon4:
@orv Curious if this is something you could use instead of your current approach? Not asking you to change what you do, but feedback would be useful.

K6AH:
Let's discuss Tuesday evening please.

ae6xe:
These new interfaces would need to be added to a network-firewall zone.  Today, mesh RF (wifi), dtdlink, lan, wan, and tun, are all firewall zones -- to update the firewall matrix of rules of what traffic is allowed from->to.    This new interface would need to have a new zone, or be attached somehow to an existing firewall zone (note that dtdlink and mesh RF are 2 zones today) to utilize on a node.

aanon4:
@ae6xe To avoid exploding the firewall rules (more than they are)  it might be simplest to add these interfaces to an existing zone. Which one makes most sense to you? The VPN zone used by tunnels might be the most similar?

ae6xe:
Before my time, in the earliest days of bbhn/aredn, the dtdlink firewall zone was created as this interface went up/down on a live node if the admin plugged in/out the cat5 cable.   The tun zone was added, with the assumption we needed a 3rd zone (we didn't question this approach at the time).   But now that it is getting more complicated, we should review the firewall rules to identify if indeed the routing across all 'mesh' like zones (wifi, dtdlink, and tun) have any specialized restrictions -- why do we need multiple zones to begin with?       Possibly, some of these dynamic interfaces are problematic if they go up/down and combined with other interfaces in the same zone -- breaking things for a brief amount of time.  At present, when a tun or dtdlink interface goes up/down, hotplug (or vtun scripts) add/remove the firewall rules to add/remove the new zone.     We might have or desire unique rules for tun, to block some traffic, e.g. routing to a WAN gateway across the tunnel.   If we create a custom interface to have different behavior (LQ/NLQ unique calculation and cost) than dtdlink, tun, or RF, then we may want to also restrict some traffic uniquely.   We may need to conf call to discuss this complexity to determine a path forward.

aanon4:
> Before my time, in the earliest days of bbhn/aredn, the dtdlink firewall zone was created as this interface went up/down on a live node if the admin plugged in/out the cat5 cable. The tun zone was added, with the assumption we needed a 3rd zone (we didn't question this approach at the time). But now that it is getting more complicated, we should review the firewall rules to identify if indeed the routing across all 'mesh' like zones (wifi, dtdlink, and tun) have any specialized restrictions -- why do we need multiple zones to begin with? Possibly, some of these dynamic interfaces are problematic if they go up/down and combined with other interfaces in the same zone -- breaking things for a brief amount of time. At present, when a tun or dtdlink interface goes up/down, hotplug (or vtun scripts) add/remove the firewall rules to add/remove the new zone. We might have or desire unique rules for tun, to block some traffic, e.g. routing to a WAN gateway across the tunnel. If we create a custom interface to have different behavior (LQ/NLQ unique calculation and cost) than dtdlink, tun, or RF, then we may want to also restrict some traffic uniquely. We may need to conf call to discuss this complexity to determine a path forward.

This seems like an orthogonal issue to where, as it is currently designed, these interfaces should be added. I've put them in the vpn (tunnel) zone for now.

Perhaps reworking this is best done in the openwrt update where we have to rewrite all this stuff to replace iptables anyway? Not much point in doing it twice.

ae6xe:
In openwrt UI and design, the interfaces are added into zones, then the rules are created between the zones.  Yes, we don't exactly do that, but we build on top of openwrt and may be making it more complex than it needs to be.   Agreed, we should revisit this in the openwrt upgrade to new firewall implementation.

ae6xe (commented):
Need to define routing for the new interface -- policy tables.   See the /usr/local/bin/vtun*  scripts:

root@AE6XE-haplite-test:~# ip rule
0:      from all lookup local
20010:  from all iif eth1.2 lookup 29
20010:  from all iif wlan1 lookup 29
20010:  from all iif tun60 lookup 29
20020:  from all iif eth1.2 lookup 30
20020:  from all iif wlan1 lookup 30
20020:  from all iif tun60 lookup 30
20080:  from all iif eth1.2 lookup 31
20080:  from all iif wlan1 lookup 31
20080:  from all iif tun60 lookup 31
20090:  from all iif tun60 lookup main
30010:  from all iif br-lan lookup 29
30020:  from all iif br-lan lookup 30
30090:  from all iif br-lan lookup main
30099:  from all iif br-lan lookup 31
30210:  from all lookup 29
30220:  from all lookup 30
30290:  from all lookup main
30299:  from all lookup 31
32766:  from all lookup main
32767:  from all lookup default

aanon4:
I've added a hotplug script which reuses the vtun infrastructure.

ae6xe (commented):
This should be equivalent --  Since vtun is no longer a separate installable package, now bundled in, opportunity to put everything in 11-meshrouting?  consolidate and simplify.

K6AH:
I'd like to understand what the use cases for this might be and why we would want to allow them.  Let's talk about it this evening.

ae6xe:
related enhancement request:  https://github.com/aredn/aredn/issues/54

aanon4:
> I'd like to understand what the use cases for this might be and why we would want to allow them. Let's talk about it this evening.

We should discuss, but let me do a bit of background writing here as well.

What is this for?

This is to let groups run AREDN networks over  non-AREDN hardware. There are probably other uses, but this is specifically to allow the larger AREDN network to use alternate radios and third-party networks. In the Bay Area we are already building out a backbone network using higher end radios which TDMA. This hardware cannot run AREDN for various reasons. Often it's newer radios where we have no support. Other times we're sharing a network which has already been built (for example, we share space on the fire monitoring network in Lafayette) and they dont run AREDN. However, these "backbone" networks allow us to increase our footprint into communities we otherwise couldnt reach.

Why this approach?

This is the absolute minimum feature set I could come up with to allow external networks to run AREDN. Essentially all it does it provide another "pipe" into OLSR and AREDN. How that "pipe" works once it leaves our nodes it deliberately undefined. All that's required is that AREDN traffic comes in and out of that pipe, just like a tunnel.

Why not just use the tunnels then?

The tunnels we have expect to run over a WAN interface. The SoCal folk "bodge" this a bit by making a node's WAN actually be another non-AREDN wireless network. This works, but its hardly elegant, and it restricts what they can build. Also, they're forced to run the tunnel protocol when there's no need adding unnecessary overhead and *much* slower network speeds.

Why not just connect all the VLAN2s together?

That's what we have now and it's not a very scalable solution. We have a network which is beginning to span the whole Bay Area, from Mount Diablo down to Monterrey Bay. Making everything pretend its on a single VLAN (think local ethernet) is very unscalable and fragile long term and it makes redundancy and failure tolerance harder to include. It also doesn't play well with some potential partners.

Can you do this another way?

I'm sure we could. But I'd rather not depend on custom firmware on "backbone" nodes. SoCal use to do this and I've been on a crusade to remove all those requirements and get everyone onto one firmware release.

aanon4:
@ae6xe I consolidated everything into 11-meshrouting

aanon4:
Thank you all for the discussion tonight - looks like we good for this feature to be included.

aanon4:
Also - for when @ae6xe or @dman776 see this - finished testing tonight and all looks good. Doesnt break anything when not in use, and works as expected when enabled.

ae6xe (approved):
Concensus was to move forward.   Looks good.

Orv:
So this can be closed now?

PR #516, opened 2022-10-08
Force a GC before the lua manager sleeps

aanon4:
Garbage collect before we sleep to keep the memory footprint down. We're about to wait anyway, so now is a good time to do this.


Orv (approved):
Looking forward to testing this!

ab7pa (approved):
Looks good.

dman776 (approved):
Looks good

PR #518, opened 2022-10-11
Fix limiting nr of route shows with low memory

aanon4:
Freemem was in bytes (not kbytes) and hash removal was broken.

### DIFF
--- a/files/www/cgi-bin/mesh
+++ b/files/www/cgi-bin/mesh
@@ -159,9 +159,8 @@ end
 local lowmemory = cursor:get("aredn", "@meshstatus[0]", "lowmem")
 if not lowmemory then
     lowmemory = 10000
-else
-    lowmemory = tonumber(lowmemory)
 end
+lowmemory = 1024 * tonumber(lowmemory)
 local lowroutes = cursor:get("aredn", "@meshstatus[0]", "lowroutes")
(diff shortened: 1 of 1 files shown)


files/www/cgi-bin/mesh
```diff
@@ -159,9 +159,8 @@ end
 local lowmemory = cursor:get("aredn", "@meshstatus[0]", "lowmem")
 if not lowmemory then
     lowmemory = 10000
```
ab7pa, line 161: Probably need to convert this value to bytes also.

aanon4:
@ab7pa The code changes looks like this:
```
local lowmemory = cursor:get("aredn", "@meshstatus[0]", "lowmem")
if not lowmemory then
    lowmemory = 10000
end
lowmemory = 1024 * tonumber(lowmemory)
```
So both values are x 1024.

ab7pa:
Got it!  I missed that you moved the conversion out of the if-then like it had been before.  Thanks!

ab7pa (approved):
Tested this with a variety of values on the Advanced Configuration -- Memory Thresholds page.  Everything worked exactly as designed!  Thanks, @aanon4 Tim.

dman776 (approved):
good stopgap

PR #519, opened 2022-10-11
add warning message to package display

ab7pa:
If a node has low free memory then display a warning message that installing extra packages may not be a good idea.

### DIFF
--- a/files/usr/lib/lua/aredn/utils.lua
+++ b/files/usr/lib/lua/aredn/utils.lua
@@ -702,14 +702,14 @@ function isFreeMemoryLow()
     -- get low memory threshold
     local u=uci.cursor()
     local lowmemory = u:get("aredn", "@meshstatus[0]", "lowmem")
-    if not lowmemory then 
-        lowmemory = 10000 
+    if not lowmemory then
+        lowmemory = 10000
     end
     lowmemory = 1024 * tonumber(lowmemory)   -- convert to bytes
-    if nixio.sysinfo().freeram < lowmemory then 
-        return true 
-    else 
-        return false 
+    if (nixio.sysinfo().freeram + nixio.sysinfo().bufferram) < lowmemory then 
+        return true
+    else
+        return false
     end
 end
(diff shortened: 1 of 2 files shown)


aanon4:
If we're going to start doing this check in various places, we need to add a function in one of our common libraries to do it rather than replicating this bit of code.

ab7pa:
Excellent!  I'll work on that.  THX

ab7pa:
Added a utility and referenced it in /www/cgi-bin/admin and /www/cgi-bin/mesh.

aanon4:
Could you be more "generous" with the code formatting rather than smashing the if..then..end statements onto single lines?

ab7pa:
Will do.  I also noticed that I wrongly included changes from your other PR in mine, so I'm going to remove the stuff I included.  Sorry about that.

aanon4:
Thanks!

aanon4 (approved):

ae6xe (commented):
Not unique to this PR...   We should consider moving to "getAvailMem()" (not sure if I have the right named function) instead of Free.   Free memory includes RAM that could be immediately freed up.  Available is a higher value than Free and is the actual memory available if a a process is starting up.   The kernel will consume RAM for performance reasons that can be freed up at any time.  Available does not count RAM that can instantly be freed up.

ab7pa:
The purpose of the utility in this PR is to test against the _Low Memory Threshold_ entered on the _Advanced Configuration_ page.  It's somewhat of a unique use case, and it is only as useful as the value that is entered there.  This is the same logic used for "low memory route reduction" on the _Mesh Status_ page.

aanon4:
I can't find "getAvailMem" - any pointers?

ab7pa:
Decision is **not** to compare the relative Free Memory value with the Low Memory Threshold value, but instead to print a generic warning message on several pages where features could have additional memory impact (packages, tunnels, etc).  Closing this PR as unnecessary.

PR #520, opened 2022-10-11
Display hosts totals rather than olsr totals

aanon4:
Count hosts and nodes using the olsr_host file rather than asking olsrd for some json.
This saves memory (and maybe some time) and gives very similar (but not identical) values. I changed the web page labels to reflect the change.

### DIFF
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -42,7 +42,6 @@ require("aredn.utils")
 local html = require("aredn.html")
 require("uci")
 local aredn_info = require("aredn.info")
-local olsr = require("aredn.olsr")
 
 -- helpers start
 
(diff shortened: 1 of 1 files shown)


ab7pa (approved):
Works exactly as designed.  Thanks, @aanon4 Tim!

ae6xe (commented):
hostnames named "mid[\d+]\.*" should be excluded.   If not excluded, this would inflate the host_total count.

aanon4:
So I tried excluding mid-s and the total count was cut by 1/3 ... which was dramatically smaller than the olsr_total count (the current code give a closer number), so I removed it. Happy to put it back, but be aware it will change the number dramatically.

ae6xe:
The prior count may be inflated too?   Definitely, the mid-s are duplicate entries.  To confirm, check a live example and sort by IP address -- every mid is a duplicate IP address of another entry.

aanon4:
Okay - no longer counting hostnames beginning mid<nr>.

ae6xe (approved):
looks great!

PR #525, opened 2022-10-14
Fix new mac extraction code in LQM was breaking for tunnels

aanon4:


ab7pa (approved):
This fixed the errors on my tunneling + RF node.  Thanks, @aanon4 Tim!

dman776 (approved):

PR #527, opened 2022-10-18
add search capability to Mesh Status page

ab7pa:
Add the ability to filter some of the key columns for _Current Neighbors_ and _Remote Nodes_ based on search characters entered in the column header fields.

### DIFF
--- a/files/www/cgi-bin/mesh
+++ b/files/www/cgi-bin/mesh
@@ -407,23 +407,47 @@ if nixio.fs.stat("/tmp/node.history") then
     end
 end
 
+------------------
+-- generate page
+------------------
 http_header()
 html.header(node .. " mesh status", false)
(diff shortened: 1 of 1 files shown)


ab7pa:
Getting some tester feedback .... stay tuned.

ab7pa:
Testers suggested making some of the key column headers filterable, so the headers are now searchable.

Screen Shot 2022-10-18 at 12 26 44 PM

dman776 (approved):

mathisono:
Why not add a "Services Name" search?  

Seeing the "This Node" info on the main "status" page would eliminate the need for myself to go to the "Mesh Status" page alot. The "Mesh Status" page could then focus on the "current neighbor" area... merged with the "Neighbor" data on the lqm page.  ???

PR #528, opened 2022-10-20
Handle names missing .local.mesh

aanon4:
Accidentally triggered this today, which breaks the mesh display. Ignore these badly formed entries.


ab7pa (approved):
Tested and works as designed.  Thanks, @aanon4 Tim, for discovering and fixing this use case!

dman776 (approved):

PR #533, opened 2022-10-23
Make ntp update script protocol agnostic

aanon4:
Some people define their ntp clocks using the pseudo protocol ntp:// which which we should allow


ab7pa (approved):
Works as designed.  Allows any protocol specification in the NTP URL.

dman776 (approved):

PR #534, opened 2022-10-24
Line up the columns on the mesh display

aanon4:
Line up the columns on the mesh display.

### DIFF
--- a/files/www/cgi-bin/mesh
+++ b/files/www/cgi-bin/mesh
@@ -443,6 +443,43 @@ html.print([[
     input.search {
         width:150px;
     }
+    #nSubTable {
+      width: 80%
+    }
+    #nTable th:nth-child(1) {
+      width: 31.5%;
+    }
+    #nTable th:nth-child(2) {
+      width: 25%;
+    }
+    #nTable td:nth-child(2) {
+      width: 43.5%;
+      padding-left: 5px;
+    }
+    #nTable td:nth-child(3) {
+      padding-left: 5px;
+    }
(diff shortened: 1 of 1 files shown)


ab7pa:
I like the columns lined up with each other, but I'm not thrilled with the _OLSR Entries_ and _Previous Neighbors_ taking up space on the right side.  Can't we just delete those two sections from the **Mesh Status** page?  IMHO they really don't add any value, but I'd like to understand why they're required if others think they're essential.

aanon4:
Screen Shot 2022-10-24 at 8 59 29 AM

aanon4:
So people can see what the proposed layout looks like with this change. Happy to remove the OLSR part, but the Pervious Neighbors need to be somewhere on this page still.

ab7pa:
Some Current & Remote node names are now wrapping if they are too long.  Also, there are some node names with a `dtdlink.` prefix that could be filtered off -- and tactical names have a `.local.mesh` suffix that could be filtered off.

ab7pa:
Example of prefix/suffix and wrapping

ab7pa:
As long as the nodenames are reasonably short the Previous Neighbors width should work.  When the nodenames get too long then they will definitely wrap.
Screen Shot 2022-10-25 at 11 56 58 AM

ab7pa (approved):
Nice to line up the columns on all tables.  Works as designed.  Thanks, @aanon4 Tim.

dman776 (approved):

PR #536, opened 2022-10-25
add kernel MemAvailable to Status page

ab7pa:
Add the _MemAvailable_ value from `/proc/meminfo` to the **Status** page.  This value is an estimate of how much memory is available for starting new applications. The linux kernel calculates this value from MemFree, SReclaimable, the size of the file LRU lists, and the low watermarks in each zone.  [See this link for more information on MemAvailable](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773).

This PR also removes the buffer cache value from the free memory calculation, which inflates the free memory value (per Joe AE6XE).  It continues to display the sysinfo value for comparison purposes.  It also removes the confusing /tmp free value, since it is not separate storage space but is already included in the available memory figure.

### DIFF
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -103,15 +103,16 @@ end
 function get_memavail()
     local f = io.open("/proc/meminfo", "r")
     if f then
-      for line in f:lines()
-      do
-        local memavail = line:match("^MemAvailable:%s+(%d+)%s+kB")
-        if memavail then
-          f:close()
-          return tonumber(memavail)
+        for line in f:lines()
+        do
+            local memavail = line:match("^MemAvailable:%s+(%d+)%s+kB")
+            if memavail then
+                f:close()
+                return memavail
+            end
         end
-      end
     end
+    f:close()
 end
 
 function get_wifi_signal(wifiif)
@@ -413,10 +414,14 @@ else
     rspace = rspace .. " KB"
 end
 local mavail = get_memavail()
-if mavail < 500 then
-    mavail = "<blink><b>" .. mavail .. " KB</b></blink>"
+if mavail then
+    if tonumber(mavail) < 500 then
+        mavail = "<blink><b>" .. mavail .. " KB</b></blink>"
+    else
+        mavail = mavail .. " KB"
+    end
 else
-    mavail = mavail .. " KB"
+    mavail = "unknown"
 end
 
 col2[#col2 + 1] = "<th align=right valign=top><nobr>load average</nobr><br><nobr>available space</nobr></th><td>" .. string.format("%.2f, %.2f, %.2f", sysinfo.loads[1], sysinfo.loads[2], sysinfo.loads[3]) .. "<br><nobr>flash = " .. fspace .. "</nobr><br><nobr>/tmp = " .. tspace .. "</nobr><br><nobr>memory = " .. rspace .. "</nobr><br><nobr>kmemavail = " .. mavail .. "</nobr></td>";
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -100,6 +100,20 @@ function get_default_gw()
     return "none"
 end
 
+function get_memavail()
+    local f = io.open("/proc/meminfo", "r")
+    if f then
+      for line in f:lines()
+      do
+        local memavail = line:match("^MemAvailable:%s+(%d+)%s+kB")
+        if memavail then
+          f:close()
+          return tonumber(memavail)
+        end
+      end
+    end
+end
+
 function get_wifi_signal(wifiif)
     local signal = -1000
     local noise = -1000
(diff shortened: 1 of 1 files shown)


files/www/cgi-bin/status
```diff
@@ -100,6 +100,20 @@ function get_default_gw()
...
+    local f = io.open("/proc/meminfo", "r")
+    if f then
+      for line in f:lines()
+      do
+        local memavail = line:match("^MemAvailable:%s+(%d+)%s+kB")
+        if memavail then
+          f:close()
+          return tonumber(memavail)
+        end
+      end
```
aanon4, line 113: Need to also close the file here - if somehow we didn't find MemAvailable we still need to close the file

files/www/cgi-bin/status
```diff
@@ -392,15 +406,21 @@ if tspace < 3000 then
...
 end
-local rspace = (sysinfo.freeram + sysinfo.bufferram) / 1024
+local rspace = sysinfo.freeram / 1024
 if rspace < 500 then
     rspace = "<blink><b>" .. rspace .. " KB</b></blink>"
 else
     rspace = rspace .. " KB"
 end
+local mavail = get_memavail()
+if mavail < 500 then
```
aanon4, line 416: While extremely unlikely, get_memavail can return nil (it didnt find a value) and this comparison will error and stop the page being rendered. Need to handle that case or return a default value from get_memavail.

aanon4 (changes requested):
A couple of things to fix. Highly unlikely theyd ever happen, but ...

files/www/cgi-bin/status
ab7pa: Good catch, @aanon4 Tim.  THX

files/www/cgi-bin/status
ab7pa: Good catch, @aanon4 Tim.  THX

files/www/cgi-bin/status
```diff
@@ -100,6 +100,21 @@ function get_default_gw()
...
+        for line in f:lines()
+        do
+            local memavail = line:match("^MemAvailable:%s+(%d+)%s+kB")
+            if memavail then
+                f:close()
+                return memavail
+            end
+        end
+    end
+    f:close()
```
aanon4, line 115: I'm sure it was my comment placement, but the f:close should be inside the if ... end not after it.

aanon4 (changes requested):

files/www/cgi-bin/status
ab7pa: Doh.   Thanks again Tim

aanon4 (approved):

ae6xe:
Possibly another discussion and PR, or option to include in this PR.   Showing free /tmp space is confusing -- it is in RAM.   Free /tmp space is already shown with the memory free and available values.    It ca be confused that this is separate space available to use -- it is not.      Also, this temp filesystem "tmpfs" grows and only consumes RAM for the actual files-data written to /tmp.   A 'df' command will show more space in this file system than physically exists.  At this moment on my hap ac lite:   /tmp "Available" space is 29760.   However,  MemFree:  20508 and MemAvailable:  28084, are less than /tmp thinks it has -- not actually available.   Suggest we remove /tmp space from the reporting, it's already included in RAM memory.

ae6xe:
I did more research on why we are seeing MemFree higher than MemAvailable.     The MemAvailable value is determined by subtracting  out some amount of unused RAM.   The amount removed is based on a  low watermark threshold where adding more pages in RAM would trigger other pages to be swapped out to disk (virtual memory).   This isn't relevant to AREDN because we don't have any swap space defined.    Is there a support dump available on the ar300m where we see this condition?   The amount subtracted out on this device just smells too high for what the low water mark should be.  Possibly, there is more to the story.

ab7pa:
[proc-info.txt](https://github.com/aredn/aredn/files/9875297/proc-info.txt)

ab7pa:
Will convert to Draft until Joe has a chance to look at this metric in more detail.  THX

ae6xe:
Suggest to move forward and use the larger value of:  "unused RAM" (free memory) or "MemAvailable".      In the scenario where MemAvailable is less than Free memory, since we do not have swap space, we should not be subtracting out an amount of RAM to prevent swapping to calculate MemAvailable.   This approach to use the larger, will be comparable and more meaningful across devices.

ab7pa:
Sounds like a good plan, @ae6xe Joe.  I will work on that tomorrow.  Many thanks!

ab7pa:
This code will now display the greater of either freeram or memAvailable.  Tested on 128mb and 64mb nodes.

ae6xe (approved):
Looks good.

PR #538, opened 2022-10-27
fix current neighbor service search and add new search fields 11/01/2022

ab7pa:
Fix service search for Current Neighbor table and include search fields for all tables.
This PR also puts Previous Neighbors at the end of the Current Neighbors table.

Screenshot 2022-10-27 at 1 54 27 PM

### DIFF
--- a/files/www/cgi-bin/mesh
+++ b/files/www/cgi-bin/mesh
@@ -706,7 +706,6 @@ else
 end
 
 --add previous neighbors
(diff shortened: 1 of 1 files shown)


Orv (dismissed):
Works as advertised.  Better....

Orv (approved):

dman776 (approved):

PR #544, opened 2022-11-04
set min field widths on ports page

ab7pa:
Set some minimum field widths on the **Ports** display. This helps with readability on various web browsers that tend to auto-squash field widths. Tested on chromium, safari, and firefox. Made the DHCP Hostname field a little wider so more characters in longer hostnames can be seen without scrolling right.

Screenshot 2022-11-04 at 2 21 07 PM


Orv (approved):
Tested - works as expected.

dman776 (approved):

PR #548, opened 2022-11-14
remove reference to nodes reducing power

ab7pa:
Remove reference to nodes reducing power levels per @ae6xe Joe.

> Re: node may reduce power at higher link rates.  While this behavior is stated in Ubiquiti specifications, I've never seen anything in ath9k or openwrt forums that suggests this occurs in the open source implementations. I'd speculate that open source, and thus AREDN, does not have this behavior. I've also not seen any information of what other vendor's do.

### DIFF
--- a/files/www/help.html
+++ b/files/www/help.html
@@ -255,7 +255,7 @@ interfaces are set.</p>
 <ul>
 <li><a name=channel></a>The <strong>Channel</strong> and <strong>Channel Width</strong> selection determines the center frequency and signal bandwidth.  AREDN&#174; reminds operators that they must select frequencies, bandwidths, and power levels which comply with their country's amateur radio license requirements.<br><br></li>
 
-<li><a name=power></a>The <strong>Power</strong> setting controls the maximum power the unit may transmit. Some devices may have max power levels that change based on which channel/frequency the hardware is operating on, and in this case the max level will change when you save settings and will be capped at the max level supported by the hardware for that frequency.</li>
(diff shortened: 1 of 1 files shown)


Orv (dismissed):
I concur.

ae6xe (commented):
This is unique to TP-Link devices.    Suggest to state accordingly in the documentation.   An inspection of /etc/radios.json will show the power max allowed by channel number, is only on TP-Link devices.

ae6xe:
"the max level will change when you save settings".   Are there examples of this?   The power value is set, then saved, then after reboot, the value changes?    This condition should be a defect -- we can have the power values, as defined by the manufacture  in the vendor flash area, set in /etc/radios.json .   We should be able to correct this data file so this symptom does not occur.

Orv (approved):
Better.

ae6xe:
"the max level will change when you save the settings".    I recall what this issue was about now.   In the perl UI implementation, when a new channel is selected, the max power field is not automatically updated.     If during a save setting, there is a power value above the defined max allowed for the channel, an error should be given -- to never save a bad value.      The condition needs to be fixed, and we would never need a readme to explain the situation (other than known bugs in the release notes).   The significant question, "do we have this error checking code in today's implementation"?   Ether we need to auto-update the max power, based on channel selection, or we need a check on "save" to report the condition and the user corrects.

ab7pa:
So this should be a separate GitHub issue, rather than something having to do with the on-node help text being updated in this PR?

ab7pa:
So the change in the on-node help file should be ready to approve now.  THX

dman776 (approved):

PR #550, opened 2022-11-15
Make sure c6 always has a value

aanon4:


aanon4:
FIxes an occasional layout failure

dman776 (approved):

PR #552, opened 2022-11-22
add WAN wifi ssid and snr to Status page

ab7pa:
Add a line showing the WAN wifi SSID and SNR if the node is using its Wifi as WAN connection.

**Here is the Status page with no WAN connection:**

**Here is the Status page with Ethernet WAN connection:**

**Here is the Status page with Wifi WAN connection:**
Screenshot 2022-11-22 at 12 49 29 PM

### DIFF
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -157,6 +157,8 @@ local config = aredn_info.get_nvram("config")
 if config == "" or nixio.fs.stat("/etc/config.mesh", "type") ~= "dir" then
     config = "not set"
 end
+
+-- get wifi interface info
 local wifi_iface = aredn.hardware.get_iface_name("wifi")
 local wifi_nr = wifi_iface:match("wlan(%d+)")
 local wifi_disabled = true
@@ -166,11 +168,15 @@ if wifi_nr then
     radio = "radio" .. wifi_nr
 end
 
+-- get wan interface info
+local wan_iface = aredn.hardware.get_iface_name("wan")
+
 local cursor = uci.cursor()
 
 local wifi_channel
 local wifi_chanbw
 local wifi_ssid
+local wan_wifi_ssid
 if not wifi_disabled then
     wifi_channel = tonumber(cursor:get("wireless", radio, "channel"))
     if wifi_channel >= 76 and wifi_channel <= 99 then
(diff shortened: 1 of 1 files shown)


Orv (dismissed):
Works as advertised.  A nice little tweak to the UI:

image

ab7pa:
I have not seen Orv's sig level on any of the nodes I've tested, but I'm putting this PR into draft mode until I have time to track that down.  The function being used to get the sig/noise for WAN Wifi is the same one being used for Mesh RF Wifi `get_wifi_signal(wifiif)`, and it looks like it's executing `iwinfo.nl80211.assoclist(wifiif)` within that function.

ab7pa:
I am able to reproduce the issue.  Every so often `iwinfo.nl80211.assoclist(wifiif)` will drop the minus sign on the signal level.  When that happens then the absolute value of the difference between signal and noise will jump up.  Refresh again and it will return to expected levels.  Not sure why this is happening, but will keep investigating.

ab7pa:
I have a workaround tested and ready to go.  However, as the saying goes, "You can't really solve a problem until you know what caused it."  And I still don't know why the minus sign is being dropped intermittently.  Will continue investigating.

ab7pa:
**get_wifi_signal()** is a carryover from the legacy Perl code.  I created a separate program using **get_wifi_signal()** to report `|signal|noise|snr|` every second, and I ran it continuously to watch the behavior of _iwinfo_.  I did not see any issues when _iwinfo_ reported metrics for the **meshRF/wifi** interface, but when reporting metrics for the **WAN Wifi** interface the `station.signal` value was missing its sign unpredictably once every 10 to 20 seconds.  The attached text file has a snippet of the results.  [iwdata.txt](https://github.com/aredn/aredn/files/10085688/iwdata.txt).  As a workaround for this condition I include a test before the return in **get_wifi_signal()** to add the sign if it is missing.  This has resolved the issue on all of my test nodes.

ae6xe:
Suggestion, but increases scope...   so let me know if you prefer to separate out of this PR, otherwise this looks ready to commit to the repo, looking good.

There are really 2 gateway paths that can be in use at the same time to make clear:   1) gateway of local LAN devices is always the WAN IP address;  otherwise  2) gateway of the mesh (traffic routing through the node) to the nearest gateway (which might be the local WAN IP address if it is advertised).

ab7pa:
@ae6xe Joe, I see what you mean and think it would be good to show both.  Probably best to work on that in a separate PR, just so I don't get confused during testing on different types of nodes to verify that it will work.  This PR has been tested on single- and dual-radio nodes with config values in various stages of completeness, etc, and it appears to work okay.  (although I'm sure the wider user community will find other issues in the NB that I haven't thought of yet)

ae6xe (approved):
looks good.

Issue #553, opened 2022-11-25
Display both possible gateway addresses on the Node Status page


ab7pa:
Suggestion from @ae6xe Joe:  There are really two gateway paths that can be in use at the same time: 

1. the gateway of node LAN devices is always the WAN IP address (if it exists);
2. otherwise it is the nearest gateway on the mesh (traffic routing through the node) which might be the local WAN IP address if it is advertised.

We should display both possible gateways on the _Node Status_ page.

ab7pa:
Completed.

PR #554, opened 2022-11-25
display both types of gateway on Status page

ab7pa:
Based on the suggestion from @ae6xe Joe ([see Issue#553](https://github.com/aredn/aredn/issues/553)) here is a possibility for displaying both types of gateway on the **Node Status** page.  Here are some samples of the display under various conditions.

**Mesh node without WAN connection:**
meshRF-noWAN

**Mesh node with wired WAN connection:**
meshRF-wiredWAN

**Node using its radio for Wifi WAN:**
noMesh-wifiWAN

**Dual radio node with Mesh RF and LAN AP:**
meshRF-lanAP

### DIFF
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -321,7 +321,7 @@ if wifi_disabled then
     col1[#col1 + 1] = "<th align=right><nobr>primary address:</nobr></th><td>" .. ip .. " <small>/ " .. cidr .. "</small><br>"
 else
     wifi_gw = get_default_gw("wifi")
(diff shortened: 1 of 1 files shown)


ab7pa:
Some suggestions from testers: eliminate the extra section labels and possibly add information when in LAN AP mode.  Putting into draft mode to take a look at doing these things.

Orv (approved):
Final version works nicely, is more compact, but still readable.

ae6xe:
There might be a way to clarify who/what is using a given gateway (for the 2 paths).   Instead of "default gateway", this might be "LAN device's gateway".   Instead of "mesh gateway", if it will fit could be clearer to say "mesh traffic gateway" (or "mesh traffic gateway path").   the mesh gateway is an IP address of the next mesh node in the direction towards the gateway -- the actual gateway may be 4 hops away, not the next hop.   Not sure if these are the best terms, but putting it out on the table of the issue to see if we can come up with optimal names.

aanon4:
"mesh traffic gateway path" ? I dont think we want these labels to become descriptions - users should read the manual if you want to know more.

aanon4 (approved):

Orv (approved):

PR #556, opened 2022-12-01
Fix Bad Gateway error when trying to update with the wrong firmware

aanon4:


ab7pa (approved):
Looks good, @aanon4 Tim!

dman776 (approved):

PR #557, opened 2022-12-05
update vtun name when changing nodename

### LINKED ISSUE #555: [Bug] Tunnel settings are not refreshed when renaming a node
While doing network maintenance jobs, I needed to rename a distant tunnel client node to reflect a site callsign change. The node name has been also changed on the tunnel server side. Both from the GUI.

After the client node has been rebooted, the tunnel will not go up. On both side, we can see the connection is rejected.

I logged in into the client node, edited /etc/config/vtun and /etc/config.mesh/vtun to change the callsign portion of the "name". Rebooted the client and now the tunnel will go up.

---

Changes needed: When the node name is changed, the tunnel "name" values for tunnel clients must also be changed. Those values could be computed at every reboot to make it simple.

ab7pa:
Update the tunnel client name in /etc/config/vtun & /etc/config.mesh/vtun when a Tunnel Client node is being renamed.  Fixes #555

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -567,6 +567,15 @@ if not ntp_period then
     ntp_period = "daily"
 end
 
+function is_vtun_client()
+    for line in io.lines("/etc/config/vtun") do
+        if line:match("^config server") then
+            return true
+        end
+    end
+    return false
+end
+
 -- validate and save configuration
 if parms.button_save then
     for _,zone in ipairs(tz_db_names)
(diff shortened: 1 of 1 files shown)


files/www/cgi-bin/setup
```diff
@@ -742,6 +751,35 @@ if parms.button_save then
...
+                        for k,v in pairs(section) do
+                            if k == ".name" then
+                                servnum = "@" .. v:gsub("%_","%[") .. "]"
+                            end
+                        end
+                        local newname = parms.node .. "-" .. section.netip:gsub("%.", "%-")
+                        if newname and servnum then
+                            cursor:set("vtun", servnum, "node", newname)
+                        end
+                    end
```
aanon4, line 776:
```suggestion
                    function(section)
                        for k,v in pairs(section) do
                            if k == ".name" then
                                local servnum = "@" .. v:gsub("%_","%[") .. "]"
                                local newname = parms.node .. "-" .. section.netip:gsub("%.", "%-")
                                cursor:set("vtun", servnum, "node", newname)
                                break
                            end
                        end
                    end
```
This is a little cleaner.

aanon4 (changes requested):
A made a few suggestions on the code (but I admit I didnt actually test them - sorry)

PR #559, opened 2022-12-06
fix status page not rendering

### LINKED ISSUE #558: Mikrotik HAP ACL doesn't render entire first screen below the buttons N-1948 to N-1955
Some (not all) Mikrotik HAP ACL will not render the entire first GUI screen below the Buttons.  All other screens appear to be fine.  So far this has occurred on 3 nodes out of probably more than 15 that have been upgraded. Same behavior Google Chrome, Microsoft Edge, Palemoon 32 and Firefox.  Mesh RF is disabled seems to be the key.  See below.

ab7pa:
Reproduced Issue #558 and provided a fix for that issue as well as the unclosed _nobr_ tag. Fixes #558

### DIFF
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -284,7 +284,7 @@ html.print("&nbsp;&nbsp;")
 html.print("<select name=\"css\" size=\"1\" onChange=\"form.submit()\" >")
 css_options()
 html.print("</select>")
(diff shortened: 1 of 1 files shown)


ab7pa:
The last two testers commented here that this fixed the issue: https://github.com/aredn/aredn/issues/558

PR #561, opened 2022-12-10
Check /etc/config/vtun exists

aanon4:
This files doesnt exist when node is installed, cuases a bad gateway failure and stops the node being setup.


ab7pa (approved):
Very good catch.  This should go in right away.  THX

dman776 (approved):

PR #573, opened 2022-12-21
fix local mesh status bug

### LINKED ISSUE #571: Service names don't line up with hostnames on Mesh Status page
In the first section of the Mesh Status page, the service names are listed all together without line breaks so they do not line up with their respective hosts. The same list shows up correctly when viewed on remote nodes. Cosmetic issue only.

3.22.12.0

image

ab7pa:
Fix #571 -- alignment of local nodes mesh status display.


dman776 (approved):
good catch

PR #596, opened 2022-12-27
Fix package dependency update

aanon4:
New okpg reports missing dependencies differently, so handle this case.


dman776 (changes requested):
This needs to split into 2 PRs as they are different topics.

aanon4:
Looks like I didn't base it correctly as those other changes are already submitted. Will kill this one and redo.

PR #598, opened 2022-12-27
separate devices from standard README

ab7pa:
Make it easier to find Supported Devices by separating from the more static README file, making the README file much easier to maintain going forward.


aanon4:
Also, if we end up generating this SUPPORTED_DEVICES file from some other data source, much cleaner to do it like this.

aanon4 (approved):

dman776 (approved):

PR #605, opened 2022-12-29
new SUPPORTED_DEVICES.md format

dman776:
proposed new format for discussion

### DIFF
--- a/SUPPORTED_DEVICES.md
+++ b/SUPPORTED_DEVICES.md
@@ -40,10 +40,10 @@ Bullet M5 | 5 | ath79 | - | - | 32MB | unknown | sunset
 Bullet M5Ti | 5 | ath79 | - | - | 32MB | unknown | sunset
 Bullet M2 | 2 | ath79 | tiny | ubnt_bullet-m-ar7241 | 32MB | stable | sunset
 Bullet M2 XW | 2 | ath79 | generic | ubnt_bullet-m-xw | 64MB | unknown | released
-LiteBeam AC5 Gen2 | 5 | ath79 | generic | ubnt_litebeam-ac-gen2 | 64MB | stable | new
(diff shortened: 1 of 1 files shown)


VA2XJM:
For machine reading uses OR markdown cut/paste, the status column should not be left empty. It should indicate something like "Actual" or "Active" or "Current".

VA2XJM:
It should also be markdown formated as a table, It would make it a very handy format. I missed that on first pass.

ab7pa:
Could we change the "new" _Status_ to "nightly" and then in the _Notes_ say that this device is supported only in the nightly build firmware at this time?  For a device with _Stability_ of "untested" maybe we could say _Stability_ is "unknown" and Status is "untested"?  What would you think of eliminating the Meraki device, since it is not supported in these Nightlies or in future Stable releases?

dman776:
it *IS* a markdown table now...
For standardized terms, what do you think about:

- Stability: "Unknown", "Stable", "Bricked" and/or "Unstable"
- Status: "Released", "Nightly"

This way, when we prepare the "release" PR, one step would be to simply change all the "Nightly" to "Released" in this file.

ab7pa:
Sounds good to me. Good idea.  And the MD table looks fine to me.

aanon4:
Also seems good to me.

ab7pa:
I think these GL.iNet devices are "unknown" and "nightly" only.
```
GL.iNet GL-AR300M w/ 128MB NAND | 2 | ath79 | nand | gl-ar300m | 64MB | unknown | released
GL.iNet GL-AR750s (slate) | 2 | ath79 | nand | gl-ar750s | 128MB | unknown | released
GL.iNet GL-E750 | 2 | ath79 | nand | gl-e750 | 128MB | unknown | released
```

ab7pa (approved):
I think this looks good.  Thanks, Darryl
Is there any way to display this table in Drupal (arednmesh.org) exactly as it's being displayed here in GitHub?  The current Drupal version is very difficult to read.

PR #613, opened 2023-01-03
Fix negative channels not beaconing.

aanon4:
Also update patch for -3 and -4 channels.


aanon4:
The majority of this patch is the update to bring in the release patch which supports channels -3 and -4. There is one tiny fix which makes the -ve beaconing work where some math was being done using unsigned ints, but because the channel is negative, this was giving the wrong value.

dman776 (approved):

PR #615, opened 2023-01-04
Enable ac neg channels

aanon4:
This makes -ve channels work on AC devices.
This is not a perfect fix as it can still take an AC device a couple of minutes to connect to an existing network as
the initial scans fail to find it. This can probably be improved so connect is more immediate.


ae6xe (approved):
Looks good.

PR #616, opened 2023-01-04
More ways to force HX as the default country.

aanon4:


ae6xe:
For the use case where 1 interface is configured HX (mesh RF) and the 2nd is, e.g. US (LAN AP) -- shouldn't defaulting to any value always be over-ridden and must work?    If the per interface config is not working, then there is always a situation where the default will be wrong--meaning this change might not resolve the issue.   (Part of our challenge is that we may be the only group using multiple reg settings at the same time and no one has ever tested this?)      I don't see this PR will hurt, so I'm not aware of any issues to push through.  But, just not sure if it will make a difference and this PR benefits.  Thoughts?

dman776 (approved):

ae6xe:
Let me dig a little.  There are 3+ regulatory settings (global, phy0, phy1, ...) as shown by "iw reg get".    I do see that "iw reg set" help doesn't show an interface in the option...  Maybe it is a matter of digging through the UCI commands for the country config option to see what it is doing?

root@AE6XE-haplite-test:/lib/wifi# iw reg get
global
country HX: DFS-UNSET
        (2300 - 2600 @ 40), (N/A, 40), (N/A)
        (5170 - 6500 @ 40), (N/A, 40), (N/A)

phy#1
country US: DFS-FCC
        (2400 - 2472 @ 40), (N/A, 30), (N/A)
        (5150 - 5250 @ 80), (N/A, 23), (N/A), AUTO-BW
        (5250 - 5350 @ 80), (N/A, 23), (0 ms), DFS, AUTO-BW
        (5470 - 5730 @ 160), (N/A, 23), (0 ms), DFS
        (5730 - 5850 @ 80), (N/A, 30), (N/A), AUTO-BW
        (5850 - 5895 @ 40), (N/A, 27), (N/A), NO-OUTDOOR, AUTO-BW, PASSIVE-SCAN
        (57240 - 71000 @ 2160), (N/A, 40), (N/A)

phy#0
country US: DFS-FCC
        (2400 - 2472 @ 40), (N/A, 30), (N/A)
        (5150 - 5250 @ 80), (N/A, 23), (N/A), AUTO-BW
        (5250 - 5350 @ 80), (N/A, 23), (0 ms), DFS, AUTO-BW
        (5470 - 5730 @ 160), (N/A, 23), (0 ms), DFS
        (5730 - 5850 @ 80), (N/A, 30), (N/A), AUTO-BW
        (5850 - 5895 @ 40), (N/A, 27), (N/A), NO-OUTDOOR, AUTO-BW, PASSIVE-SCAN
        (57240 - 71000 @ 2160), (N/A, 40), (N/A)

aanon4:
Well this change only forces the initial state, so if there's a way it get's changed to US, BR, UK, FR or whatever when the AP is active, it should still work if there's an api for it.

ae6xe:
I'm seeing that the country code is passed to hostapd from the /etc/config/wireless settings.    I'd speculate that hostapd will then set the correct country code on the interface going into AP mode.    IBSS mode doesn't seem to have an equivalent of hostapd to be setting the country code on the given interface.  I agree, let's proceed -- don't have a better option to propose.

ae6xe (commented):
looks good.

PR #618, opened 2023-01-05
lqm: Switch from wget to curl for better control of timeouts

aanon4:
meshchat uses curl which has much better control over timeouts for bad connections. Nick that code and use it here also.


ae6xe (approved):
looks good.

PR #620, opened 2023-01-05
Fix tunnel firewall

aanon4:
Data destined for the firewall endpoint was being blocked by bad rule ordering.


ae6xe (approved):
Looks good.

PR #622, opened 2023-01-06
Tag devices which must be reinstalled.

aanon4:
Looks like the NAND layout for a few Mikrotik devices has changed sufficiently that they cannot be easily upgrade and must be reinstalled from scratch.  Note this in the SUPPORTED_DEVICES matrix.


ae6xe (commented):
"this device is supported, but cannot be upgraded and must be reinstalled."   Propose something to the effect of, and relevant for multiple future releases:  "Upgrades from 3.22.12.0 and before, can not directly use a sysupgrade image/process, rather must start as if replacing vendor firmware with the initramfs image, then use sysupgrade image."    The statement as is, would not apply, e.g. to 3.24.01.0 when upgrading from 3.23.06.1.

aanon4:
I took another stab at the language while still trying to keep it terse and general.

ae6xe (approved):
Looks good.

PR #627, opened 2023-01-07
Remove LQM first run code

aanon4:
This was used during the transition to using LQM and is no longer needed.


ae6xe (approved):
Looks good.

PR #628, opened 2023-01-08
Fix idle tunnel quality check 01/07/2023

aanon4:
When a tunnel is idle, binding to the tun* device fails; so remove it. As we have a direct tunnel route in the routing table (not OLSR table 30) created by vtun, we will still correctly route the quality testing traffic.
Validated for both idle and active tunnels.


ae6xe (approved):
Looks good.

PR #629, opened 2023-01-08
Add a maximum timeout for service checks 01/08/2023

aanon4:
The connect timeout doest not include DNS lookups, and if DNS is broken this can hang forever. Add a maxmimum timeout so this call will eventually terminate regardless.


ab7pa (approved):
Looks good.  Nice that curl provides the option.

ae6xe (approved):
good find-catch.  Looks good.

PR #635, opened 2023-01-11
Hidden node detection (and optional mitigation) 01/12/2023

aanon4:
A new LQM feature which detects hidden nodes which a node cannot see directly and display them as part of the LQM information. It can also (but is off by default) enable RTS/CTS to improve quality by reducing transmission failures when hidden nodes are detected.

RTS/CTS mitigation is off by default, even when hidden nodes are detected. But when it's enabled I've been able to do a little testing locally on a few Omni's we have locally, and its reduced transmission failures from 30% to about 8%. I would expect more, but in my test only a single Omni was using RTS/CTS while all its neighbors were not.

### DIFF
--- a/files/etc/uci-defaults/80_aredn_lqm
+++ b/files/etc/uci-defaults/80_aredn_lqm
@@ -7,6 +7,7 @@ do
         option enable '0'
         option min_snr '15'
         option margin_snr '1'
+        option rts_theshold '256'
         option min_distance '0'
         option max_distance '80467'
         option min_quality '50'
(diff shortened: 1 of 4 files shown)


Orv:
Need to include some documentation someplace how to manage this feature.

aanon4:
True enough, although I'd like some feedback on the how its displayed in the Neighbor page before we do that in case people have idea and changes.

ae6xe (approved):
Looks good.

aanon4:
So wifi radios are alway good citizens - is someone sends an RTS to another radio, that radio will reply with a CTS and everyone who hears it will respect it, regardless of whether anyone else has it enabled. So CTS is *always* enabled, its just a question of whether RTS is used to trigger it. In the tower example, the tower will respond with a CTS as it should, but unless it has RTS enabled, it wont use that before transmitting itself.

PR #641, opened 2023-01-12
Block older versions of the firmware once we upgrade

aanon4:
You cant (easily and reliably) go back once upgrading this time, so block older firmware versions from being offered.


ae6xe (approved):
Looks good.

PR #650, opened 2023-01-15
New network configuration code

aanon4:
I'd hoped to avoid doing this ... but here we are ...

With some of the new devices come new network setups internal to the devices. This is particularly true of the new Mikrotik AC devices. The older way of configuring VLANs will no longer work. Somewhat problematically it appears to work, but unfortunately it leaks mac address between vlans and physical networks, causing lots of problems.

OpenWRT introduced the 'bridge-vlan' as the new way to configure vlans on bridges, and these work for both older devices as well as new ones. The changes to use them are quite extensive, and to keep things sane I took the opportunity to pull all the network and wifi configuration code into 'node-setup' where some pieces were already. 

The code has been tested on my 22 node bank (including old 32M devices, 3G and new AC devices) as well as some by-hand testing on the hAP, hAP AC2 and AC3 and the AR150.


ae6xe (approved):
Ok

ae6xe:
Some of the 2-port XW and XM devices have internal switches.  They have been configured with Linux thinking there is a single interface and no vlans.   This PR, still leaves this same configuration on these devices?   We've never been able to have tagged and untagged on the same port, and this approach has been the only known way these devices could support thus standard configuration.

aanon4:
Have you seen a problem? I've tested this on XM and XW devices and not seen an issue, but I may have missed something.

ae6xe:
These devices override the default vlan settings from openwrt.   If
02_network is reverted back to openwrt, then we'd still need the aredn
custom changes.   To test, make sure the secondary port can do LAN and
dtdlink.   Both ports need to test LAN use at same time.   I'm out of town
and no access to equipment for now.

PR #653, opened 2023-01-15
Generalize node-setup variable expansion

aanon4:
Move the node-setup variable expansion into a function so we can reuse it.


ae6xe (approved):
Looks good.

Issue #633, opened 2023-01-10
AP doesn't boot with active WAN


KenArck:
Nightly 2124-6dac4c6 

hAP AC2 - if booted (cold or warm) with WAN connection made, AP doesn't broadcast SSID so can't connect. If booted  (cold or warm) without WAN connection, AP comes alive

[supportdata-AH6LE-HAP-SHOP-202210191622.gz](https://github.com/aredn/aredn/files/10377901/supportdata-AH6LE-HAP-SHOP-202210191622.gz)

aanon4:
Thanks. Did this work okay on a previous nightly (it's probably a firewall change)?

KenArck:
It worked as of 3.22.12 (the last version I flashed to my hAP Lite). It only started happening as of the first nightly build made available for the hAP AC2 ( I believe I first placed it into service on Monday, Jan 3rd). I've since reflashed with build I posted in my OP

aanon4:
Can you describe what's plugged into you hAP AC2? Is there just the single cable plugged into the WAN port? If not, what else is connected. Do these cables plug into switches? Are any external VLANs configured?
Thanks.

KenArck:
WAN port is plugged directly into TPLink AX1800 Router. Port 1 connected to TPLink TL-SG108E switch. Port 2 connected to GrandStream GXP1620 phone

aanon4:
Could I request the following while running the current nightly:
1. Boot without WAN - wait 10 minutes - check the status of the AP - send support data.
2. Boot with WAN - wait 10 minutes - check the status of the AP - send support data.
Thanks!

KenArck:
Only problem is it will not boot with WAN hence I can't download the support file

aanon4:
I have my dev hAP AC2 on my desk (it has a serial cable connected which is helpful at times). If I boot it up with just a WAN cable connected it does boot. For me when I insert the power, the PWR light comes on, then in a moment the USR light comes on indicating the device is in the boot loader. After a while this light goes off again, then starts to flash as the AREDN software boots. Eventually it goes steady again as boot is completed.  Very occasionally I've noticed inserting the power never has the USR light come and the device doesn't correct power up (no output on the serial line). This hasn't seemed correlated to what cables I have inserted. I've found unplugging for a dozen seconds the retrying helped (immediately retrying didn't).

[...]

PR #654, opened 2023-01-16
Mikrotik boot loader change to avoid boot lockup problem

aanon4:
https://github.com/aredn/aredn/issues/633

Reports that the hAP AC2 will fail to boot when the WAN cable is connected. This does not effect all devices. However, I've also seen this on my hAP Lite device and it is not a new problem (as long as I've owned it!). The forum post below indicates the boot loader can be set to the wrong mode, causing the device to lockup sometimes when booting. Checking my two hAP AC2, they were both in different modes. It looks like it might be possible to accidentally change this mode by holding the reset button for different periods of time when powering on the device. This might explain why some nodes have this problem and some don't.

See: https://forum.openwrt.org/t/openwrt-installed-mikrotik-rb750gr3-wont-boot-when-wan-cable-plugged/146330/7

### DIFF
--- a/files/etc/uci-defaults/12_mikrotik_boot
+++ b/files/etc/uci-defaults/12_mikrotik_boot
@@ -1,0 +1,11 @@
+#! /bin/sh
+#
+# On Mikrotik devices, make sure the boot mode is 'flasheth' otherwise
+# the node can fail to boot.
+#
+if [ -f /sys/firmware/mikrotik/soft_config/boot_device ]; then
+    if [ "$(grep '\[flasheth\]' /sys/firmware/mikrotik/soft_config/boot_device)" = "" ]; then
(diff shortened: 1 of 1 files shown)


ae6xe (approved):
Looks good.

aanon4:
Keeping it in draft until the reported confirms this fixes the issue.

aanon4:
Fantastic. @ae6xe you may need do something to merge this as the approval came before I switched it out of Draft.

Orv:
On my ac2:
Contents of /sys/firmware/mikrotik/soft_config/boot_device:  eth [flasheth] ethonce flash cfg cfgonce 
Contents of  /sys/firmware/mikrotik/soft_config/commit:  clean

Are those adequate?

dman776 (approved):

aanon4:
@Orv That's what you want them to look like - yes.

PR #656, opened 2023-01-17
Fix power offsets

aanon4:


aanon4:
Will update this as I test devices and convert from draft when I'm done.

dman776 (approved):

PR #658, opened 2023-01-19
Cleanup the initramfs images for everything but Mikrotik

aanon4:


aanon4:
Ah .. you did mention that and I forgot. I'll kill this and look elsewhere for a fix.

PR #659, opened 2023-01-19
Automatically enable RTS when hidden nodes detected

aanon4:
As we though this was on anyway, I'd say we should enable RTS whenever we detected hidden nodes in order to establish a more reliable network. It can still be disabled in the advanced settings.

Thoughts?

### DIFF
--- a/files/etc/uci-defaults/80_aredn_lqm
+++ b/files/etc/uci-defaults/80_aredn_lqm
@@ -7,7 +7,7 @@ do
         option enable '0'
         option min_snr '15'
         option margin_snr '1'
-        option rts_theshold '-1'
+        option rts_theshold '1'
         option min_distance '0'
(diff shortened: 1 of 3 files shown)


dman776:
looks reasonable to me

apcameron:
Looks good to me

K6AH:
Seems reasonable to me also.

dman776 (approved):

Orv:
One more time - what's the definition of a hidden node, and how does the node know about it?   (Asking for a friend - seriously - Steve wants to get it right in the docs :-)  )

aanon4:
A hidden node is a node which is communicating with one of your RF neighbors, but you cannot see.
How does it know? ... there's code in LQM which works it out ... what do you need here specifically?

PR #660, opened 2023-01-19
Initramfs cleanup

aanon4:
Take two. I tidied the Makefile, but the actual removal is done by the changes in the config files so the various json files are correct.


dman776 (commented):
cant we just discard the hunk for the Makefile change as it should be unchanged from the last commit?

dman776:
and thank you for doing this the proper way!

dman776 (approved):
looks great

Issue #663, opened 2023-01-20
STABLE experience report (aredn-2195-6aa2903-ath79-generic-ubnt_nanobeam-ac-xc-squashfs-factory)


cTurtle98:
this is a report of my experience with nightly `2195` on the `nanobeam-ac-xc`

the device can be listed as stable in the snapshot supported devices list

--------

DETAILS:

```
ciaran@ciaran-ThinkPad-P14s-Gen-2a:~/Downloads$ scp aredn-2195-6aa2903-ath79-generic-ubnt_nanobeam-ac-xc-squashfs-factory.bin   admin@192.168.1.20:/tmp/factory.bin
Warning: Permanently added '192.168.1.20' (DSA) to the list of known hosts.
admin@192.168.1.20's password: 
aredn-2195-6aa2903-ath79-generic-ubnt_nanobea 100% 5441KB   3.9MB/s   00:01    
ciaran@ciaran-ThinkPad-P14s-Gen-2a:~/Downloads$ ssh admin@192.168.1.20
admin@192.168.1.20's password: 
 ___  ___
|\  \|\  \      Welcome to airOS
\ \  \\\  \
 \ \  \\\  \    PRODUCT: NanoBeam 5AC 19
  \ \  \\\  \	MAC: 44D9E7485C87
   \ \_______\
    \|_______|  VERSION: XC.v8.7.0

XC# hexdump -Cv /bin/ubntbox | sed 's/14 40 fe 27/00 00 00 00/g' | hexdump -R > 
/tmp/fwupdate.real
XC# chmod +x /tmp/fwupdate.real
XC# /tmp/fwupdate.real -m /tmp/factory.bin
Writing 'kernel         ' to /dev/mtd2(kernel         ) ...  [%100]
Writing 'rootfs         ' to /dev/mtd3(rootfs         ) ...  [%100]
Done
```
got the nocall home at `192.168.1.1`

![image](https://user-images.githubusercontent.com/24465356/213808630-8e338c21-6fbb-4536-8f57-76830d1a7a17.png)

set the hostname and the password

![image](https://user-images.githubusercontent.com/24465356/213808805-f6a3e2ce-f0e8-4723-bb6c-3890ed479f62.png)

click save and get the reboot notice

![image](https://user-images.githubusercontent.com/24465356/213809063-8a4988b0-3c29-46e8-8852-68b44c97199f.png)

[...]

PR #661, opened 2023-01-20
Update the status of some devices which have been tested

aanon4:
Thanks to K1KY

PR #664, opened 2023-01-20
change nanobeam-ac-xc stability to stable

cTurtle98:
issue #663


cTurtle98:
> This is contained in #661

if you approve this PR then I get to be in the contributors list...
that would be cool for me
you can close it if you would prefer

dman776:
Already updated with #661

PR #665, opened 2023-01-21
Ignore non-routable when calculating hidden nodes

aanon4:


dman776 (approved):

Orv:
What's defines a non-routable node?   Users are asking :-)

aanon4:
Is there an active route in the routing table to that node - ie. will any traffic be sent on that RF path.

Issue #637, opened 2023-01-11
Package Management shows ~40 options to download 'Package version' in 2131 build


ae6xe:
on Mikrotik ac2, after refreshing the 'download packages', there were ~40 selection entries to download 'package version'.  The actual package names are not showing, except 'kmod*' and 'block-mount*'.   I was looking for tcpdump to help debug some issues.  On 2131 nightly build image.  The refresh occurred before the next nightly build.

dman776:
may be specfic to the ipq40xx architecture.    It works find on my hAP AC Lite (generic/mikrotik)

dman776:
what are the values in your AdvConfig firmware paths?   are they correct for the arm_cortex..... packages?

dman776:
it’s hard coded in the advandedconfig page in the defaultPackageRepos() function around line 77
need to pull that value from “/etc/openwrt_release DISTRIB_ARCH” instead

ae6xe:
Confirmed, all the paths on the hap ac2 need to have changed in the download paths:  from "mips_24kc", to "arm_cortex-a7_neon-vfpv4".

dman776:
fixed by #672

PR #672, opened 2023-01-25
Make sure we look for packages with the correct architecture

aanon4:
https://github.com/aredn/aredn/issues/637


ae6xe (approved):
Looks good.

PR #673, opened 2023-01-25
Remove firewall counters except for specific ports

aanon4:
I think I added these before I understood you didnt have to.


ae6xe (approved):
Looks good.

Issue #567, opened 2022-12-15
Nightly Builds Download


kc2cbd:
Is it possible to add a Nightly Builds Download Box on the Administration Page?

dman776:
not currently.   it would involve the manipulation of all of the ".list" files EVERY TIME a nightly build is run.

dman776:
looking into this further, more logic would be required because the "snapshots/nightlies" are stored in a different directory on the downloads server than the released firmware.  (including needing to know the architecture and subtype for the firmware.  ie.  /snapshots/targets/ath79/generic vs /snapshots/targets/ar71xx/mikrotik vs /firmware)

PR #675, opened 2023-01-25
New firmware selector

aanon4:
Based on discussions last night, here a new firmware selector which uses the same information as the firmware selector UI.

Advantages:
* Uses the same information, so no need to generate our own special info files (well .. one last time when we release)
* Includes the ability to download the current nightly
Disadvantages:
* Now has to make 1+n network requests to populate the firmware versions dialog (previously just 1 request). The first request gets the available versions, but these are not all valid for all hardware, so extra requests are necessary to validate these.

Leaving it here for comment and further testing.

### DIFF
--- a/files/www/cgi-bin/admin
+++ b/files/www/cgi-bin/admin
@@ -43,6 +43,7 @@ aredn.html = require("aredn.html")
 require("uci")
 aredn.info = require("aredn.info")
 require("ubus")
+require("luci.jsonc")
 
 local html = aredn.html
 
@@ -51,10 +52,9 @@ local conn = ubus.connect()
 
 -- handle firmware updates
 local fw_install = false
-local patch_install = false
 local fw_output = {}
-local fw_images = {}
-local fw_md5 = {}
+local fw_versions = {}
+local fw_names = {}
 local fw_version = ""
 local blocked_fw = {
     "^aredn%-3%.15",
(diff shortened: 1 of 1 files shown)


dman776:
Wow!   Tim, that was fast!  i love it.
i loaded this on a hap ac lite.   it only shows the nightly build and not the recent releases.  perhaps due to the rename of the hardware type (ie.  rb-xxxx to routerboard-xxxx)?

aanon4:
Yes, it wont show the older generic names for devices. On the other hand, once you upgrade you can't go back to them anyway as the names wont match the folder firmware so it'll refuse to flash it.

aanon4:
Which we need to remember when messaging around the release because unlike others, a lot of devices can't go back.

dman776:
partially fixes #567

aanon4:
Do people want to push this into the nightly? Any concerns?

ae6xe (approved):
Looks good.

k6ccc:
I like the ability to download the current nightly from the Download firmware selector - even though I don't normally use it.  That will make it easier for people to install / test the nightly builds.
For whatever it's worth, I loaded this NB onto 14 nodes.  After the update, as expected many types did not show anything but NB 2231 as available.  The only nodes that did show the 3.22.x.x release versions in addition to NB 2231 were my three Rocket M3 (XM) and two Rocket M5 (XM) nodes.  These types did not show anything except NB 2231:  Rocket M5 (XW), hAP AC Lite, LHG 2nD, LHG 5nD, LHG 5HPnD, AR750, USB150.

aanon4:
Thanks for checking this - very much appreciated. As you say, is should only show versions you can actually load - and as you can't go back from this nightly to early versions on many radios now (due to internal name changes) what you see sounds correct.

k1ky:
I have utilized this feature on at least the last 3 nightly releases and works as advertised.  Nice feature!

mathisono:
Worked great today...!  

Should I expect the next stable release to only be pushable via file upload to devices running a nighty build?

aanon4:
When the next release comes out, you'll be able to upgrade in all the usual ways from both current nightlies as well as the previous release.

Issue #674, opened 2023-01-25
Nightly 2224 (and previous) limit TX Power to 25 dBm


nc8q:
sysinfo:
 node: NC8Q-Kettering-Centerville
model: MikroTik RouterBOARD LHG 5nD
status:
firmware version: 2224-4894f42
model: MikroTik RouterBOARD LHG 5nD
This is actually a MikroTik RouterBOARD LHG 5HPnD-XL and is being limited to 25 dBm.

sysinfo:
MikroTik RouterBOARD LHG 5HPnD-XL
 node: NC8Q-Centerville-Kettering
model: MikroTik RouterBOARD LHG 5HPnD-XL
status:
firmware version: 3.22.12.0
model: This is correct and allows 28 dBm

sysinfo:
 node: NC8Q-Centerville-HuberHeights
model: MikroTik RouterBOARD RBLHG-5HPnDstatus:
status:
firmware version: 3.22.12.0
model: MikroTik RouterBOARD RBLHG-5HPnD
This is correct and allows 28 dBm

sysinfo:
 node: NC8Q-HuberHeights-Centerville
model: MikroTik RouterBOARD RBLHG-5HPnDstatus:
firmware version: 1827-2711589
model: MikroTik RouterBOARD RBLHG-5HPnD
This is correct and allows 28 dBm

vy 73, Chuck

ae6xe:
The max power settings have some constraints to consider.   As an example, on my hap ac2, the linux driver will allow a setting up to 30dBm on all channels (AREDN settings may further restrict max power).    If we refer back to the vendor specifications, the max is 27dBm on the hap ac2, and higher link rates are lower power.   See attached spec for hap ac2 vendor specs.
![hap-ac2-specs](https://user-images.githubusercontent.com/25602681/214656385-cc7c293e-62b7-4a87-ae7d-bbc52f8b5b74.JPG)

AREDN should limit the hap ac2 to 27dBm max, otherwise the signal is at risk of being distorted.  Generally this is due to cheap power amps, and avoidance to over-drive (out of linearity) this PA and bad things happen -- noise outside the channel, distorted signal, etc.   Higher link rates (vs lower link rates) can also cause the PA to be over-driven, hence the vendor specs and firmware limiting the max power further. 

[...]

PR #677, opened 2023-01-26
Split the various Mikrotik radios into their individual variants

aanon4:
Split the various Mikrotik radios into their specific variants.

Although some of these are the same software wise (XL v non-XL for example) we've moved towards specific builds for specific radios so these should follow. Also makes it more obvious when selecting firmware.


aanon4:
Amongst other things, addresses https://github.com/aredn/aredn/issues/674

dman776 (approved):

Issue #645, opened 2023-01-13
GL-iNet USB150 bricked when updated to 2145 or 2152


k6ccc:
Updated all 14 of my nodes to Nightly 2145 and all worked except my USB150.  The 2145 update was made over an RF link of about 6 feet.  When I got home and checked it out, a computer would recognize that a USB device was plugged in, but could not get an IP address from it.  Power cycled several times with the same result.  I don't remember for sure what the previous version was, but it was a nightly from sometime in the past week.

This morning used the GL-iNet unbrick process to unbrick the USB150.  Then did an update to 3.22.12.0  That properly came up with the first boot screen and then after the next reboot, it was operating properly as far as I could tell.  I don't have any other 2 GHz AREDN here to test an AREDN RF link with.  Next updated to nightly 2152.  Again, the USB150 bricked.  The attached computer recognizes that a USB device has been plugged in, but can not get an IP address from it.

Repeated the unbrick and then AREDN load and update process with the same result.

aanon4:
I'll dig out my USB150 this weekend and take a look. My guess is that something changed specifically for this device when we did the OpenWRT upgrade to their latest version earlier this week. Anyway, dont try an upgrade on the USB150 again until I update this ticket.

k6ccc:
I'm gonna leave it at 3.22.12.0 for now...

kc2cbd:
The last couple of version of nightly have been weird with the USB150. It now sort of powers down and doesn't come back up after upgrade.

k6ccc:
Update.  KC2CBD hit the nail on the head.  After the update, the node powers off and does not power back on.  However after updating, if you wait for the node to power off (about 4 minutes), then you power cycle it, and wait several more minutes, it will come back up with the new firmware and appears to be working correctly.

kc2cbd:
I was actually going to create an issue but K6CCC beat me to it.

k6ccc:
Same issue with nightly 2159.  After the update the USB150 powers down, but does not power back up.  Must be manually power cycled to get it to boot up.

k6ccc:
Still the same thing with NB 2166.  Node powers down, but does not power back up after firmware update.  Requires manual power cycle.

aanon4:
Thanks for checking. I promise to update this ticket when there's any relevant changes though ... if you want to save yourself some time.

aanon4:
Update:
[...]

PR #678, opened 2023-01-26
Add a note about the USB150 device

aanon4:


dman776 (changes requested):
it's the AR150 that I have that doesn't reboot.  I havent tried the USB150

aanon4:
Oh! I've been trying the USB150 and it doesn't reboot - so maybe it's both?

aanon4:
This was in relation to https://github.com/aredn/aredn/issues/645 which I dont think will be fixed anytime soon.

aanon4:
@dman776 do i need to add the note to the AR150 also?

dman776 (approved):

PR #690, opened 2023-02-02
Filter out non-routable ARP entries which confuse LQM

aanon4:
Very occasionally I'm seeing ARP entries for IP addresses which are not routable. These always seem to be for the IP address of a device with mesh RF disabled. It's not clear how this can happen as this IP is not assigned to the DtD VLAN which is how they're being reported in the ARP cache. Also, this problem shows up in both the current release and the nightlies. It doesn't effect the routing on the node (the routing table is correct) so these "bad" arp entries never cause anyone but LQM a problem. This change stops LQM being confused by them.

### DIFF
--- a/files/usr/local/bin/mgr/lqm.lua
+++ b/files/usr/local/bin/mgr/lqm.lua
@@ -362,18 +362,22 @@ function lqm()
         for mac, entry in pairs(arps)
         do
             if entry.Device:match("%.2$") or entry.Device:match("^br%-dtdlink") then
-                stations[#stations + 1] = {
-                    type = "DtD",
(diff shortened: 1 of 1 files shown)


aanon4:
Any pending issues here?

ae6xe:
The main IP address is reassigned to, e.g. eth0.3975 (forget the actual #) when mesh RF is turned off.  Since dtdlink is, e.g. eth0.2, the 2 interfaces have the same MAC address.    I'd expect the arp table to show the main and dtdlink IP addresses to have this same MAC address of eth0 on the node.

ae6xe (approved):
looks good.

PR #691, opened 2023-02-02
Provide a timeout on the iperf client.

aanon4:
This cryptic bit of shell script adds a maximum timeout for the iperf client to run as it appears it can get stuck occasionally. The server has a built-in timeout (not available in client mode)

### DIFF
--- a/files/www/cgi-bin/iperf
+++ b/files/www/cgi-bin/iperf
@@ -75,7 +75,7 @@ else
             f:close();
         end
         if status:match("running") then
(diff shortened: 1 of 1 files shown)


ae6xe:
This appears to kill the shell that starts iperf.   But won't that leave iperf still running and ppid now something like '1'?

aanon4:
Killing the shell takes out the entire process group of which iperf3 is a part. The pid is not 1. You can check the value if you run the code and also play around with the sleep time to verify it works as intended. Yes .. it was a pain to come up with something that would work :-(

ae6xe:
OK, so I think this works due to the fact that the shell does not have background control built in, therefore the shell is the process leader of the group.  Thus killing the shell, which is the PGID, will also kill iperf.  Otherwise, a shell with background control may not have been the process group leader (required to get iperf killed).     If it was the scenario where iperf was not killed, its PPID parent process ID would have become '1' when the shell PPID was killed, as I recall.

ae6xe (approved):
looks good.

PR #692, opened 2023-02-03
add rfband to MeshRF section

ab7pa:
Display the band in the _Mesh RF_ section on the **Setup** page, in preparation for possibly allowing Mesh RF band selection on dual radio devices.

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -1071,8 +1071,14 @@ if nixio.fs.stat("/etc/config/unconfigured") or parms.button_reset then
 end
 
 if wifi_enable == "1" then
-    html.print("<tr><td>SSID</td><td><input type=text size=15 name=wifi_ssid value='" .. wifi_ssid .. "'>-" .. wifi_chanbw .. "-v3</td></tr>")
     hidden[#hidden + 1] = "<input type=hidden name=wifi_mode value='" .. wifi_mode .. "'>"
+    local rfband = aredn.hardware.get_rfband()
+    if rfband == "900" then html.print("<tr><td>RF Band</td><td>900MHz</td></tr>")
+    elseif rfband == "2400" then html.print("<tr><td>RF Band</td><td>2GHz</td></tr>")
+    elseif rfband == "3400" then html.print("<tr><td>RF Band</td><td>3GHz</td></tr>")
+    elseif rfband == "5800ubntus" then html.print("<tr><td>RF Band</td><td>5GHz</td></tr>")
+    end
+
     html.print("<tr><td>Channel</td><td><select name=wifi_channel>")
     local rfchannels = aredn.hardware.get_rfchannels(wifiintf)
     table.sort(rfchannels, function(a, b) return a.number < b.number end)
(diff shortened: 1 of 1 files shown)


files/www/cgi-bin/setup
```diff
@@ -1087,6 +1093,7 @@ if wifi_enable == "1" then
     do
         html.print("<option value='" .. width .. "'".. (wifi_chanbw == tostring(width) and " selected" or "") .. ">" .. width .. " MHz</option>")
     end
+    html.print("<tr><td>SSID</td><td><input type=text size=15 name=wifi_ssid value='" .. wifi_ssid .. "'>-" .. wifi_chanbw .. "-v3</td></tr>")
```
aanon4, line 1096: This looks like its in the wrong place? It seems to be inside the <select> ... </select> now.

files/www/cgi-bin/setup
```diff
@@ -1071,8 +1071,14 @@ if nixio.fs.stat("/etc/config/unconfigured") or parms.button_reset then
 end
 
 if wifi_enable == "1" then
-    html.print("<tr><td>SSID</td><td><input type=text size=15 name=wifi_ssid value='" .. wifi_ssid .. "'>-" .. wifi_chanbw .. "-v3</td></tr>")
     hidden[#hidden + 1] = "<input type=hidden name=wifi_mode value='" .. wifi_mode .. "'>"
+    local rfband = aredn.hardware.get_rfband()
+    if rfband == "900" then html.print("<tr><td>RF Band</td><td>900MHz</td></tr>")
+    elseif rfband == "2400" then html.print("<tr><td>RF Band</td><td>2GHz</td></tr>")
+    elseif rfband == "3400" then html.print("<tr><td>RF Band</td><td>3GHz</td></tr>")
+    elseif rfband == "5800ubntus" then html.print("<tr><td>RF Band</td><td>5GHz</td></tr>")
```
aanon4, line 1079: This information use to be part of the title for this block, which now just says "Mesh RF" rather than including the rfband info. I took it out during the port to the latest OpenWRT because the old way of identifying this information didnt work and I was too lazy to come up with a new one. I think having it back is good, but let's put it back in the title block.

aanon4 (changes requested):
See comments at various points in the PR.

aanon4:
On the more general note about selecting the band for Mesh RF on devices where there's multiple radios, a tweak to the UI to allows selection of either 5G or 2G is good so long as we dont want to allow *both* radios for the mesh - then the UI changes become trickier.

ab7pa:
Yes, the goal of this addition was to prepare a place for a dropdown selector for Mesh RF band.  Sounds like this is not really going to work as desired, so I'm closing this PR.  THX

ae6xe:
I had this feature working in an early code branch.   Let's not loose any work in this direction...

ab7pa:
Yes, @ae6xe Joe, if you have this feature working in the **lua** _Setup_ page please add it -- it's becoming a high demand feature.  Many thanks.

ae6xe:
It was previously implement3ed in perl -- still needs to be ported to lua.

ab7pa:
@ae6xe Joe, can you provide a link to your earlier code repo where you had the prototype dual radio mesh RF working?  I'm not seeing that code anywhere.  THX

ae6xe:
it would be difficult to pull out all the relevant code.   But in /www/cgi-bin/setup, /usr/local/bin/wifi-setup, /etc/config.mesh/*, /etc/config/*, and /www/cgi-bin/perlfunc.pm, and probably other places I don't recall.   

https://github.com/ae6xe/aredn/tree/openwrt-21.02

I had removed, e.g. the "rfband" entry, now in radio.json, and {setup, wifi-setup} would detect which band the radio was, cycling though wlan0, wlan1, ... .

PR #696, opened 2023-02-03
Propose adding copy of release notes to repo

aanon4:
I'm proposing we keep a copy of the release notes in git for people who might want to easily refer to changes over many releases. I've copied the last 3 years of notes into the file. We'd update this file only when we make releases (not nightlies).

Thoughts?


aanon4:
Per discussion tonight, this seems like a good idea. I only went back 3 years here (can do more if people want) and will own updating this with each release.

dman776 (approved):

PR #697, opened 2023-02-04
Fix PowerBridge M5 network

aanon4:
PowerBridge network behaves like the NanoStation


ae6xe:
I'm looking at specs and images found on the internet.  I'm only seeing 1 Ethernet port on this device.   Where's the gap?   Are there multiple flavors, like the difference between NanoStation XM and XW flavors?  This commit is for a 2 port device.

aanon4:
Interesting. I was just looking at a PowerBridge we had here which is running older firmware and that claimed it had two ethernet ports. But it's also running nanostation software so maybe that second ethernet is fake?

PR #698, opened 2023-02-05
More consistant way to provide interface mac address in overrides

aanon4:
Provide <INTERFACE_mac> variables for network overrides - a more flexible way than using the 'include' mechanism for this.


aanon4:
Just trying to tidy this up before it becomes part of release (and a commitment).

aanon4:
This is just to tidy up the user overrides a bit for people customizing their multi-port devices.

dman776 (approved):

Issue #701, opened 2023-02-09
READ.ME file update


k1ky:
The information regarding the UBNT AirRouter Lan Ports is incorrectly stated, here are suggested descriptions:
The LAN Ports are 1-3 and DTD is on Port 4 (There is no port 5 on these devices).  
The WAN Port ON ARHP accepts POE Input and the LAN port on the Injector serves as the WAN function.
The WAN port on ARAR devices, does not accept POE input and are powered via separate 5 volt input jack.

k1ky:
Looks good 2/12/13

PR #704, opened 2023-02-09
Fix AirRouter port description

aanon4:
See https://github.com/aredn/aredn/issues/701


k1ky:
I have provided some clearer port function information in #701

aanon4:
I think the new port info I added is correct. I didnt add power information as that's not included for the other HAP-like devices.

dman776 (approved):

PR #706, opened 2023-02-10
Disable RTS by default in ath10k devices 02/11/2023

aanon4:
On the Rocket 5AC Lite if we enable RTS the device fails to send packets.
At the moment we dont understand why or what devices are effected (non-AC devices seem fine).
But disable for every device for now while a solution is found.

### DIFF
--- a/files/etc/uci-defaults/80_aredn_lqm
+++ b/files/etc/uci-defaults/80_aredn_lqm
@@ -7,7 +7,7 @@ do
         option enable '0'
         option min_snr '15'
         option margin_snr '1'
-        option rts_theshold '1'
+        option rts_threshold '-1'
         option min_distance '0'
         option max_distance '80467'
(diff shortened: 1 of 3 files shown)


aanon4:
Also .. I cant spell threshold .. apparently.

aanon4:
Some more notes on how we found this problem. With a Rocket AC radio on a hill, communication with a non-AC dish pointing at it failed when the AC radio enabled RTS. However, two AC radios point-to-point (one a Rocket) works okay, so the problem is subtler than RTS on AC's not working.

(Problems found while field testing AC radios)

ab7pa (approved):
Looks good.  THX

aanon4:
Now only for ath10k devices as we dont see this on ath9k devices.

dman776 (approved):

PR #707, opened 2023-02-10
Reworked ARP cache 02/11/2023

aanon4:
There is a problem in the arp cache where wifi ARP/IP pairs are assigned to the dtd device. Filtering these early doesnt work, so instead we filter these when rebuilding the dtdlink station list.

(Problems found while field testing some AC devices)


ab7pa (approved):
Tested fine on hAP ac lite.  THX

dman776 (approved):

PR #708, opened 2023-02-10
add extended channels to LAN AP list

ab7pa:
When configuring a node's LAN AP we already include the option to select the standard lower wifi channels.  In 2020 the FCC extended the channel list to include the [upper U-NII-3/4 channels](https://en.wikipedia.org/w/index.php?title=List_of_WLAN_channels&section=5#5_GHz_(802.11a/h/j/n/ac/ax)) for indoor use.  We are including them because some newer wifi client equipment can use those upper channels.

Screenshot 2023-02-10 at 9 39 54 AM


dman776:
Any discussion on this?  @ae6xe

ae6xe:
This shows AREDN at the front of new technology  and standards.   My vote, we move forward.    However, anticipating that some will not realize 2022 and older clients will not be compatible, some users may think it is not working to use these channels.    A release and/or help note  may help to head off support requests.

ab7pa:
Added on-node help text for Standard vs Extended channel selection.

ae6xe (approved):
looks good.

ae6xe:
oops, hit the commit too soon and didn't merge the 2 commits together...

PR #710, opened 2023-02-12
Fix AC coverage calculation in driver. 02/12/2023

aanon4:
For some reason, there was code in the driver to block the setting of the coverage when a previous setting wasn't a particular value - a value set by the same code!! It's unclear what this was trying to achieve or prevent, but it stopped AC devices operating efficiently (by a factor of 10x or more).

The coverage being wrong had the knock on effect of breaking RTS because the driver wasn't waiting the correct time when using the protocol.

### DIFF
--- a/files/usr/local/bin/mgr/lqm.lua
+++ b/files/usr/local/bin/mgr/lqm.lua
@@ -227,12 +227,6 @@ function lqm()
     -- Let things startup for a while before we begin
     wait_for_ticks(math.max(1, 30 - nixio.sysinfo().uptime))
 
-    -- Detect ath10k - we need some special handling (for the moment)
-    local is_ath10k = false
-    if nixio.fs.stat("/sys/kernel/debug/ieee80211/" .. phy .. "/ath10k") then
-        is_ath10k = true
-    end
-
     -- Create filters (cannot create during install as they disappear on reboot)
     os.execute(NFT .. " flush chain ip fw4 input_lqm 2> /dev/null")
     os.execute(NFT .. " delete chain ip fw4 input_lqm 2> /dev/null")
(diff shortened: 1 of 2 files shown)


aanon4:
Just holding this as a draft until the final testing is done. Not expecting any changes at this point.

ab7pa:
Good catch, @aanon4 Tim!  Thanks for fixing this!

aanon4:
Just finished up the RTS/CTS testing with Orv this morning and it all looks good - so taking this out of Draft.

ab7pa (approved):
Thanks, @aanon4 Tim.

dman776 (approved):
nice catch and good work

ae6xe:
The slot time for 802.11/a/b/g/n/ac is ether 9 or 20 us.    this is hard coded in the specifications:  https://en.wikipedia.org/wiki/DCF_Interframe_Space    Removing a check to find the values outside of the specification would be error handling -- shouldn't be triggered to begin with.    

What I suspect is occurring, is that when we implemented 5 and 10 MHz channel widths, this means that slot time is doubled when bandwidth is cut in half.  Consequently, this error handling doesn't know about the change in timing and was being triggered, causing the symptoms we see.   The error handling was taken out with this commit.   The test probably should change to know about the timing for smaller channel widths instead -- there's probably other reasons it was there, that may still exist.     Try a test with the original code in a 20MHz channel width.  The symptoms likely can not be reproduced.    

The bigger concern is that any hard coded timing values are suspect and need updated to account for channel width timing changes.

aanon4:
We didnt see the error at 20MHz (we checked that already) and this sounds like why that was. I'll look at how best to reimplement the check adjusting for the lower bandwidth channels - although I suppose this is just a sanity check and should never fail.

PR #712, opened 2023-02-13
Fix the bandwidth reporting for ath10k devices

aanon4:
Fix scaling the bandwidth reporting in LQM and for ath10k devices in the mesh page.

### DIFF
--- a/files/usr/local/bin/mgr/lqm.lua
+++ b/files/usr/local/bin/mgr/lqm.lua
@@ -242,6 +242,22 @@ function lqm()
     -- Or any hidden nodes
     os.execute(IW .. " " .. phy .. " set rts off > /dev/null 2>&1")
 
+    -- If the channel bandwidth is less than 20, we need to adjust what we report as the values from 'iw' will not
+    -- be correct
+    local channel_bw_scale = 1
+    local chanbw = read_all("/sys/kernel/debug/ieee80211/" .. phy .. "/ath10k/chanbw")
+    if not chanbw then
+        chanbw = read_all("/sys/kernel/debug/ieee80211/" .. phy .. "/ath9k/chanbw")
+    end
+    if chanbw then
+        chanbw = tonumber(chanbw)
+        if chanbw == 10 then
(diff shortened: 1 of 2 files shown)


ae6xe (changes requested):
This change is good and necessary.  However, there is one more change needed to account for all the rates.   802.11n has count 16 MCS rate indexes.   802.11ac has count 20.   Need to add 4 more index rates to show.  The index strings, e.g. MCS8 may not be the same rate or exist, and have changed/shifted in 802.11ac, however 16 rates are the same as 802.11n, with 4 more rates added in ac spec using 256-QAM modulation for the carrier waves.

ae6xe (commented):
hang on... for 802.11ac, this is pulling from iw to get the raw rate?   Then we don't need to update the tables, if strictly for 802.11n...

ae6xe:
If I understand the code correctly, then we are showing on:
802.11n device:   the (link rate) * (the success rate of xmit'ed frames)  =   e.g. 96% of the MCS rate
802.11ac device:  the exact MCS rate not accounting for success rate. = 100% of the MCS rate

This means on 802.11n devices the TxMbps shown will be a little less (generally) than showing the same rate and link on an 802.1ac.  

Out for comment to others.  Should we make 802.11n devices report the same as 802.11ac devices for consistency?  On 802.11ac devices, we don't have access to this rate selection and frame success data -- we only have the raw MCS rate in use.

aanon4:
The 802.11n code is unchanged.
The 802.11ac code is doing the division based on the bandwidth as the 'iw' command reports the speeds as if the bandwidth was 20MHz. The 802.11ac code doesnt look at the MCS encoding at all, but instead get the rate in Mbps as reported by 'iw'.  It does not account for error rate because that information isn't available in the 802.11ac driver.
If you want to remove the error rate accounting from the 802.11n reporting we can - let me know what you'd prefer and if so I'll make that change in a concurrent PR.

ae6xe (approved):
looks good.

PR #713, opened 2023-02-15
Ath9k - no error accounting

aanon4:
To follows https://github.com/aredn/aredn/pull/712

Calculate the tx rate the same way for both ath9k and ath10k drivers so we have consistent reporting.


aanon4:
@ae6xe This uses the same mechanism for both ath9k and ath10k devices. I'll leave it up to you whether to commit or close.

ae6xe (changes requested):
remove the rateL and rateS tables, no longer in use.

aanon4:
@ae6xe You probably need to push it again as there was a merge conflict which I fixed.

ae6xe (approved):
looks good.

PR #714, opened 2023-02-15
Force dnsmasq to update itself if no network changes for > 60secs

aanon4:
On small networks there are not a lot of OLSR name changes. While dnsmasq watches for changes and updates itself, it will sometimes miss them. On busy networks this doesnt matter as the next change will catch it up. But on smaller network (esp. test networks) a missed change can stop name resolution working for some time. So now, if no changes are detected for > 60 seconds, we force dnsmasq to reload its tables.


ae6xe (approved):
looks good.

PR #717, opened 2023-02-17
Fix fccid beacon

aanon4:
Old version (using lua socket library) no longer worked. Rebuilt using standard nixio library.
Fixed whitespace


ae6xe (approved):
Looks good

PR #718, opened 2023-02-17
Improve AP band selection.

aanon4:
Handle dual band non-hap radios where second band is 2GHz.

The whole radio selection/management needs to be revisited at some point later this year.


aanon4:
Thanks to KJ6WEG for testing this for me.

ae6xe (approved):
Looks good

PR #719, opened 2023-02-18
Fix rule checking for existing drop rules.

aanon4:
When the counters were removed from the rules, the drop checks were not updated and so were failing.
This means we got *thousands* of rules in the firewall table which, eventually, slows everything down.


ae6xe (approved):
Looks good

PR #724, opened 2023-02-19
More fixes for AP mode

aanon4:
AP radio isn't always phy1 so handle these cases.
Add in some labels to RF and AP for which bands are in use.


ae6xe (approved):
looks good.

PR #725, opened 2023-02-19
remove WAN Wifi Client key lower limit

ab7pa:
The WAN Wifi Client password is currently limited to zero characters (for connecting to an open AP) or between 8 and 64 characters.  Unfortunately the node owner has no control over the number of characters in the key for the Wifi AP he is using.  This situation occurs fairly frequently when using a travel node at a hotel.  Chuck NC8Q just ran into the issue at a Country Inn and Suites in Fredericksburg VA where the key was only 7 characters in length.  As an upper limit hopefully 64 characters will be enough, but this PR removes the lower limit on the number of characters allowed for the WAN Wifi AP key.

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -662,11 +662,11 @@ if parms.button_save then
         end
     end
     if wifi3_enable == "1" then
-        if #wifi3_key ~= 0 and (#wifi3_key < 8 or #wifi3_key > 64) then
-            err("WAN Wifi Client Password must be between 8 and 64 characters")
+        if #wifi3_key > 64 then
+            err("WAN Wifi Client Password must be 64 characters or less")
         end
         if wifi3_key:match("'") or wifi3_ssid:match("'") then
-            err("The WAN Wifi Client password and ssid may not contain a single quote character")
+            err("The WAN Wifi Client password and ssid cannot contain a single quote character")
         end
     end
 


aanon4:
Does Chuck NC8Q know anything more about this situation? While the spec for WPA keys is 8-63 characters, implementations are certainly a little more hazy. I think some version of Linux are okay with 128 character keys. The smallest key I saw was 5 characters, which is what WEP uses, so I was wondering if is was possible that Chuck's hotel was using WEP?

Anyway, I dont have an issue with removing the lower limit as long as it doesnt crash the AP daemon. Specs are fine until the real world gets in the way.

ab7pa:
@nc8q Chuck, can you provide any details?  THX
The AP daemon would be running on the hotel's equipment, but I wouldn't think our short keys would crash it.

aanon4:
Sorry - you're right, the AP is on their end - I meant the WPA supplicant bit on our end. I'm sure the worst case is that it just wouldnt work.

nc8q:
I modified  'setup' as Steve suggested and I got  connected to the motel's wifi. Thanks.  Chuck.Sent from my Galaxy
-------- Original message --------From: Tim Wilkinson ***@***.***> Date: 2/19/23  4:57 PM  (GMT-05:00) To: aredn/aredn ***@***.***> Cc: nc8q ***@***.***>, Mention ***@***.***> Subject: Re: [aredn/aredn] remove WAN Wifi Client key lower limit (PR #725) 
Sorry - you're right, the AP is on their end - I meant the WPA supplicant bit on our end. I'm sure the worst case is that it just wouldnt work.

ae6xe:
per openwrt:
=============
In any WPA-PSK mode, this is a string that specifies the pre-shared passphrase from which the pre-shared key will be derived. The clear text key has to be 8-63 characters long. If a 64-character hexadecimal string is supplied, it will be used directly as the pre-shared key instead. In WEP mode, this can be an integer specifying which key index to use (key1, key2, key3, or key4.) Alternatively, it can be a string specifying a passphrase or key directly, as in key1. In any WPA-Enterprise AP mode, this option has a different interpretation.
=============
We have tested that 7-char is working, but it is unclear if upstream openwrt "wifi" commands will fail when no min 8 per documentation.    Shall we proceed as is?  I don't have an alternative to propose.

ab7pa:
Yes, AREDN correctly enforces the 8 - 64 char limit on its **LAN AP** key.

This PR is about allowing zero to 64 chars for a **WAN Wifi Client** key which has been set on a completely separate non-AREDN wifi AP that the node owners are trying to link with in order to obtain WAN Internet access for their node.  We already allow zero characters to connect with open external wifi APs.

ae6xe (approved):
looks good.

PR #728, opened 2023-02-20
Fix AP mode selection when turning mesh back on.

aanon4:
When re-enabling the mesh radio, the phy will be nil at that moment (but not after a refresh) - handle this case.


ae6xe (approved):
looks good.

PR #739, opened 2023-03-03
Added nanobeam-m5-19

aanon4:


ae6xe:
This is an "8xx" range patch for a persistent AREDN patch classified as "other".   Wouldn't this be a "0xx" patch and added upstream in the future, we only need this in the short term?  Ref:  https://github.com/aredn/aredn/blob/main/patches/PATCHES.rst

aanon4:
I can move it into the 7xx range, but it needs to come after 749. If I move it back to 0xx then we'll have to rebuild a large number of patches as they'll no longer apply.

ae6xe:
If these changes are added upstream, then won't it be the same issue of updating AREDN persistent patches that apply on top?  It's just a question of taking the hit now or later?  Happy to push through now, is that what you want to do?

aanon4:
If we could push this now that'd be good. And I will add it to my list of things to do after the next release.

ae6xe (approved):
Looks good.

PR #742, opened 2023-03-07
iperf: Improve error reporting when server is busy/disabled

aanon4:
Error reporting, especially when using the api interface, wasn't very helpful especially if iperf on the server was disabled.


ae6xe (approved):
good enhancement.

PR #745, opened 2023-03-09
Support forced upgrades

aanon4:
This has come up a few times as people have started to use nightlies and have ended up on the wrong one (having assuming they should continue to use an incorrect model version because that's what they use to do). The fix is to allow technically incompatible upgrades - technically because they might be perfectly valid .. but there's no easy way to know. So, here's a mechanism which requires a push of a button in the advanced configure, and show a big red banner in the firmware upgrade UI. The trade off here is that providing this mechanism with large red warnings will be less of a support problem compared to people using this to shoot themselves in the foot.

UI for setting the option:

Screenshot 2023-03-08 at 10 25 25 PM

Banner show when option is set:

Screenshot 2023-03-08 at 10 26 04 PM

### DIFF
--- a/files/usr/local/bin/aredn_sysupgrade
+++ b/files/usr/local/bin/aredn_sysupgrade
@@ -49,4 +49,12 @@ case "$(/usr/local/bin/get_boardid)" in
         ;;
 esac
 
-/sbin/sysupgrade $*
+#
+# Let users force upgrade a node
+#
+force=""
+if [ -f /tmp/force-upgrade-this-is-dangerous ]; then
+  force="--force"
+fi
+
+/sbin/sysupgrade ${force} $*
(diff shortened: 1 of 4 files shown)


aanon4:
@ab7pa Would appreciate feedback on the UI part in particular.

ae6xe (approved):
I think UI language is sufficient warning.

k6ccc:
Yea!  I am happy to see this.  I have one node that somewhere along the line decided that it is a different version and now is requiring the wrong firmware.
model:  MikroTik RouterBOARD LHG 5HPnD
This is actually a LHG 5nD - not the HP version.
This is:  http://k6ccc-rb-lhg-5nd.local.mesh for those with access to the SoCal network

I assume that after loading NB 2414 (still requiring the HP version firmware), this option will be available for the NEXT update so I can force the correct version firmware.

k6ccc:
Loaded on that one "wrong" node.  Looks good!
Thanks guys!

k6ccc:
Followup on my previous.
This morning I set the "Dangerous Upgrades" to yes on the node I referred to previously and successfully loaded the LHG-5nD firmware back into my node and now it correctly thinks that it really is a LHG-5nD and not the HP version.  Thanks guys!

BTW, I like the fact that after the update it automatically set the "Dangerous Upgrades" back to OFF.  Good idea!

PR #750, opened 2023-03-13
Fix support for lhgg-5acd-xl

aanon4:


ae6xe (commented):
Wouldn't radio.json also need an update?  The model string is the same?

aanon4:
I just missed these on my first attempt sometime back. The rest was already in place.

ae6xe (approved):
looks good.

kearneyjf:
Installed Build 2423 on my MikroTik RBLHGG-5acD-XL. Solved the problem.
Thanks very much for the quick fix.
John, KM6ZJT

PR #751, opened 2023-03-14
Validate state of services over a period of time before disabling advertisements

aanon4:
Maintain state on when addresses and services have been seen and only disable them if they've not passed validation for a period of time. As configured here, a service would need to fail 3 times in a row over 2 hours before the advert is removed. The advert is always restore if a service later passes the test (just as before).

The goals is to avoid disable advert for services which might occasionally fail or be briefly offline, but just get unlucky with the validation test.

### DIFF
--- a/files/usr/local/bin/olsrd-config
+++ b/files/usr/local/bin/olsrd-config
@@ -199,7 +199,8 @@ if validate then
     for _, service in ipairs(services)
     do
         local proto, hostname, port, path = service:match("^(%w+)://([%w%-%.]+):(%d+)(.*)|...|[^|]+$")
-        if vstate[hostname:lower()] or dmz_mode == "0" then
+        local vs = vstate[hostname:lower()]
+        if not vs or vs > now or dmz_mode == "0" then
             if port == "0" then
                 -- no port so not a link - we can only check the hostname so have to assume the service is good
                 vstate[service] = last
@@ -266,7 +267,7 @@ if validate then
         if not vs then
             -- New services will be valid for a while, even if they're not there yet
             services[#services + 1] = service
-            vstate[services] = last
+            vstate[service] = last
         elseif vs > now then
             services[#services + 1] = service
         end
--- a/files/usr/local/bin/olsrd-config
+++ b/files/usr/local/bin/olsrd-config
@@ -43,6 +43,8 @@ require("uci")
 
 -- Whether to validate hosts and services before publishing
 local validate = false
+local validation_timeout = 150 * 60 -- 2.5 hours (so must fail 3 times in a row)
+local validation_state = "/tmp/service-validation-state"
 
 -- check what config gile we are building for
 local uci_conf_file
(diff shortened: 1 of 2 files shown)


aanon4:
@ae6xe hopefully this will take care of that camera which seemed to occasionally fail validation.

ae6xe (approved):
Looks good.

convolved-KY4G:
Will this affect advertised services that do not have a link?  I advertise "WINLINK" and "POST OFFICE", which are necessary in order to run WINLINK over AREDN, but have no process to link to.  I'm not sure how those two cases specifically are validated by this or the previous code.  They used to get disabled at times and no longer show up.  I have not noticed this behavior recently.  I am currently running nightly build 2414-21115f7 on GL-iNet AR300M16 nodes.

aanon4:
@convolved-KY4G This change doesnt disable more things, so if your services were not getting disabled, they still wont. However, it sounds like they were sometimes? Are they blue clickable links and if so, what happens when you click on them?

convolved-KY4G:
I wonder if they were getting disabled when I had them as blue clickable links as there is no underlying process to be called relating to Winlink (it runs on a connected computer).  That would make sense.  The links are disabled now (text is black) and probably should have been all along.  RMS Express just looks for the presence of the text "WINLINK" or "POST OFFICE" in the services list in order to recognize the capability.  If I try to enable the link, an exclamation mark shows up to the right of the line and states that the service is not being advertised (see image below).  Sometimes the text shows up as advertised on the mesh status page and sometimes it does not.  I have not been able to figure out why.  I think without the link box checked it does show up much more often.
image

convolved-KY4G:
Elaborating on my post above...
Mesh status from remote node
image

Mesh status from the winlink node (WINLINK and POST OFFICE are not appearing sometimes)
image

aanon4:
If you indicate the service is a link, and uses http, the system will look to see if it exists and talks http. And if not, it will stop publishing it. If you dont set it as a link, it will always publish the text as it's just text.

mathisono:
I cant stand the frustration ive had with this "feature" I dont need a check I need to have ANY link i add as an Advertised Services to be displayed. If reachable or not!

I have a web site! that's reachable at http://mse-87.local.mesh/ or 10.145.91.197  

What am I missing? why will the node not Advertised Services this service?

aanon4:
Probably shouldnt be debugging this on a PR which is already closed and not related to what's going on here.

But ... 

If you check the file /tmp/etc/olsrd.conf on that node you'll find that all your services are there. This validation process removes them if they're not valid, so it clearly thinks they're fine.

However ...

All the host are marked as "Do not propagate" which means the hostname will not be broadcast across the mesh, meaning they can't be found, and the links wont work (so I think OLSR just throws them on the floor). Maybe they were always marked like this, but "Do not propagate" was broken for the longest time and simply didnt do anything. To get things working again you need to propagate the hostnames.

PR #754, opened 2023-03-16
Stop RETURN from refreshing the mesh page 03/16/2023

aanon4:
When people are using the filters on the mesh page, they sometimes hit RETURN. This submit the page (which is a form) causing the page to reload, lossing the filter, and as a side effect, setting the page to auto-refresh.


ab7pa (approved):
Works perfectly.  Thanks, Tim.

ae6xe (approved):
Looks good.

PR #760, opened 2023-03-24
Improve ipFromCIDR function to handle /32 cidrs.

aanon4:
Unlike any other /X suffix, the /32 is omitted when turned into a string.


ae6xe (approved):
Looks good.

PR #762, opened 2023-03-24
Partial support for Nanobeam 2AC.

aanon4:
Known issues:
1. First device to have 2 2GHz radios which we do not support. Trying to use the second radio for AP or WAN client will fail.
2. Will only mesh at 20MHz.


aanon4:
Open to debate about whether to include this for the upcoming release. It only partially works, and at least some of the fixes require reworking the way we handle the radios, which wont happen this side of the release.

Thoughts?

aanon4:
Taking this into draft. Seems channel -2 doesnt work even at 20 Mhz (+ve channels are okay).

PR #764, opened 2023-03-25
Dont let services uses hostnames which are not propagated.

aanon4:
OLSR doesnt allow this and they wont be advertised.


ae6xe (approved):
Looks good.

PR #766, opened 2023-03-29
Force badly associated stations to reassociate.

aanon4:
There appears to be a bug in the ath10k firmware for Mikrotik devices (maybe others) where a station will occasionally associate but only broadcast traffic will be passed - unicast traffic will fail. This code detects this situation and forces the device to reassociate which fixes the problem.


ae6xe (approved):
should do the trick.

PR #767, opened 2023-03-29
General station monitor service.

aanon4:
It turns out this station bug is not limited to the ath10k driver, so make this monitor service wifi generic.
(I've now seen this at both ends of the Mikrotik AC <-> Rocket pair)


aanon4:
Apologies - I left the test running and it eventually failed at the other end so the fix needs to be more generic.

aanon4:
Hey @ae6xe @dman776 could we get this in by the weekend? It reverse the previous attempt to fix this (and this just monitors the problem). I think the previous fix is causing more trouble than it's fixing.

ae6xe (approved):
Ok.

Issue #723, opened 2023-02-18
supported platform matrix MikroTik model numbers are wrong


cTurtle98:
there are mikrotik devices listed with the wrong model numbers on the supported platform matrix
specifically the BaseBox and the mANTBox

`RB912UAG-2HPnD` is just a 2ghz bare routerboard
NOT an `BaseBox`

**`RB912UAG-2HPnD-OUT` is the model number for the `BaseBox 2`**

`RB912UAG-5HPnD` is just a 5ghz bare routerboard
NOT an `BaseBox`

**`RB912UAG-5HPnD-OUT` is the model number for the `BaseBox 5`**

`RB911G-2HPnD` is just a 2ghz DISCONTINUED bare routerboard
NOT an `mANTBox`

**`RB911G-2HPnD-12S` is the model number for the `mANTBox 2 12s`**

`RB911G-5HPnD` is just a 5ghz high power bare routerboard
NOT an `mANTBox`

**`RB921GS-5HPacD-15S` is the model number for the `mANTBox 5 15s`**

**`RB921GS-5HPacD-19S` is the model number for the `mANTBox 5 19s`**

if we do actually support those specific bare routerboards then we need to add a new model section `RouterBoard` and list those models in the appropriate frequency columns

cTurtle98:
RB921 is listed correctly on openwrt so I know that isnt an upstream issue
https://openwrt.org/toh/mikrotik/rb921

openwrt only has device pages for the 91x series routerboards themselves and not the products that contain the board.
https://openwrt.org/toh/hwdata/mikrotik/mikrotik_rb911g-5hpnd
https://openwrt.org/toh/hwdata/mikrotik/mikrotik_rb911g-2hpnd
https://openwrt.org/toh/hwdata/mikrotik/mikrotik_rb912uag-5hpnd
https://openwrt.org/toh/hwdata/mikrotik/mikrotik_rb912uag-2hpnd

that could explain those model numbers being listed wrong I suppose if we are just referencing the openwrt build profile

ae6xe:
There are many model numbers, with most of the vendors, where the motherboard is the same.  The specific packaging of a device may have further tailoring of the ending numbers of the model string, e.g. ubiquity model numbers end in "US" or "INTL" depending on how the firmware is configured from the factory.  Also, different gain antennas, but identical motherboard and identical firmware. 

[...]

PR #772, opened 2023-03-30
redo MikroTik supported_devices.md

cTurtle98:
I re-did the mikrotik section of the supported devices list

I changed the device column to be the "pretty name" of the product

I changed the SKU column to be the model number of the product

all names and skus are copy and pasted from mikrotik.com/hardware so that it is really easy to double click, right click, search with google, to find a part to buy

ps let me know if you like how I handled the bare routerboard vs basebox vs mantbox or if you think I should change it


cTurtle98:
this is to address #723

cTurtle98:
@aanon4 can you take a look at this and let me know what your think?

ae6xe (approved):
looks good.

PR #781, opened 2023-04-09
Eliminate false network rejoins using LQM information

aanon4:
Weak connection which aren't being used can look like faulty connections. We use LQM information to ignore these and avoid false rejoin events which degrade the network.


ae6xe (approved):
Looks good.

PR #782, opened 2023-04-10
Mikrotik SXT Lite2 support

aanon4:


ae6xe (approved):
Looks good.  Assume vlan settings section is already updated.

PR #784, opened 2023-04-11
Fix upgrade path for SXT 2

aanon4:


aanon4:
Add some compatibility stuff so we can upgrade from 3.22.12.0 which was an LHG version.

ae6xe (approved):
Looks good

PR #796, opened 2023-04-18
Device -> Model

aanon4:


Orv (commented):
I like it, but interested in comments from others.

dman776 (approved):

PR #798, opened 2023-04-19
Remove ar300m nand and nor builds which are causing confusion

aanon4:


ab7pa:
Does the **Supported Devices** file need to be updated with this change?

aanon4:
We never mention these images in supported devices.

dman776 (approved):

PR #801, opened 2023-04-21
Remove hardwired frequency tables and use information from the hardware

aanon4:
This is something @ae6xe had mentioned before the last release, but it touched on too many things to include then.

The goal here is to remove the various hardwired tables we have when determining what frequencies a devices can use, and instead read them from the hardware. This is a step towards us being able to use different radios for different things as it remove some assumptions about what radio must be used for what.

I've not been able to test this on a 900MHz device (I dont have one).


ab7pa:
Is the major change to radios.json going to impact the /www/cgi-bin/status display which gets the model name from that file?  Line 453: (aredn.hardware.get_radio() or { name = "unknown" }).name

aanon4:
No, that's taken care of. If you provide a name, that will be used; if not, the key will be used as the name (the library handles this).

ae6xe (approved):
Looks great!

dman776:
Anything in sysinfo.json or api affected by this?

PR #805, opened 2023-04-25
Restructure, modularize and tidy the navigation buttons and menus

aanon4:
This change restructures the menu system to make it more modular and more flexible with the goal to achieving a few things:

1. Provide general html navigation generator methods for both the user buttons and admin menus.
2. Allow menus to be disabled and hidden.
    a. This is used to disable the tunnel menus on tiny platforms which no longer ship with the vtund installed (due to space limitations).
    b. On Mikrotik devices, disable all the unusable buttons while running from RAM.
3. Improve messaging when installing from RAM disk to get people to the right place.
4. Fixup a bunch of visual alignment issues where menus would jump around slightly on the screen depending which page they were on.
5. Fix the Bad Gateway issues when the radio is not recognized.


dman776:
I like it.   Is /www/cgi-bin/nav/* the proper place for these?   Or, should it be in /usr/lib/lua/aredn/nav/*

aanon4:
@dman776 I wasnt sure where to put the nav and your place is better - I'll move it.

ae6xe:
ditto, great work!

aanon4:
Menus moved!

PR #806, opened 2023-04-25
Fix ar300m16 wan configuration

aanon4:


ae6xe:
it would be better to put this change in "708-define-aredn-ath79-networks.patch" and have all the network port changes in one place.

ae6xe (approved):
Looks good.

PR #807, opened 2023-04-26
Virtualized X86 support

aanon4:
This has been tested on Proxmox. Not tested on VMWare or on "real" hardware.


ae6xe (approved):
Great work, and sufficient to get in the builds and further testing.

PR #818, opened 2023-05-03
Fix upgrade typo

aanon4:


aanon4:
Looks like this snuck in with the x86 support merge.

dman776 (approved):

PR #823, opened 2023-05-05
Remove old PBE 100mb fix and /etc/rc.local where it was included

aanon4:
Remove the old PBE 100mb port fix as it's no longer necessary.
Move dnsmasq startup back in the procd system
Deprecate /etc/rc.local - no longer needed.


ae6xe (approved):
good, back to standard way.

Issue #825, opened 2023-05-06
GL-INET AR750 (Creta) Invalid Wi-Fi Channels for 2GHz MESH


k1ky:
Nightly 20230506 and a few previous? Invalid Wi-Fi 2GHz selection channels show 5GHz channels instead of 2.4GHz options.
C:\Users\k1kyt\My Drive\MESH_HAMNET\FIRMWARE\SCREENSHOTS\2023-05-06 16_19_49-K1KY-AR750-750-P1 setup.png
[supportdata-K1KY-AR750-750-P1-202304292228.gz](https://github.com/aredn/aredn/files/11413421/supportdata-K1KY-AR750-750-P1-202304292228.gz)

k1ky:
OK - Actually I see one issue already. My LAN Access Point is using 2GHz - should have been 5GHz, so additional 2GHz MESH sould not be allowed if already being used for Wi-Fi LAN Access Point.  The LAN Access Point (was) using 5GHz channels, not sure what changed there.  I'll try to change it and see what happens.

k1ky:
The plot thickens.  The LAN Access Point was using 5GHz channel 36, but when I checked the 2GHz MESH RF box, the LAN Access channel changed to 2GHz, so the real issue is inability to choose 2GHz MESH and 5GHz Lan Access point.

aanon4:
Could you tell me the configuration of the mesh and the AP before you did the upgrade? I'd like to setup things on my AR750, do the upgrade, and see what's messed up in the internal state.

aanon4:
Well even starting from a sensible place, it can get confused very quickly ... will take a look at this logic (which needed work anyway)

k1ky:
Definitely some weird issues here.  Should not allow both Wi-Fi Lan and MESH to use 5GHz. 
<img width="491" alt="image" src="https://user-images.githubusercontent.com/42675931/236648434-1f88d021-ae50-4310-ab99-9d3d93f978b0.png">

k1ky:
Before the upgrade, MESH RF was turned off. Lan Access Point was on 5GHz Channel 36. WAN Wi-Fi Client was previously being used on 2GHz, but that had been disabled at the time of the upgrade.

k1ky:
Same issue(s) with the Mikrotik HAP ac Lite
<img width="623" alt="image" src="https://user-images.githubusercontent.com/42675931/236649104-cb6dcda5-f2e2-4c13-80bc-9e9d0ad63c00.png">

[...]

PR #827, opened 2023-05-08
Fix wifi setup for multi-radio devices

aanon4:
Disable/enabling mesh wifi on multi-radio devices could confuse the allocation of the radios to specific tasks. This was because it was always allocating wlan0 when re-enabling, and this isn't always correct on multi-radio devices.

https://github.com/aredn/aredn/issues/825

### DIFF
--- a/files/etc/mesh-release
+++ b/files/etc/mesh-release
@@ -1,1 +1,1 @@
-KN6PLV-network-dsa-vlans-24a817a
+KN6PLV-main-a5725ee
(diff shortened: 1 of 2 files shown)


ab7pa (approved):
Tested the new **setup** file on _hAP ac lite_ and it worked fine.  Not sure of the purpose of the /etc/mesh-release change.

aanon4:
:-( /etc/mesh-release was added by accident. I'll remove it.

ae6xe (approved):
looks good.

Issue #19, opened 2019-09-06
DTD SWCONFIG Options by GUI - Feature Request


k1ky:
Enhancement request to be able to view port configuration(s) and  change the DTD VLAN configuration of ports on AirRouters,  HAP and compatible GLInet  and similar future network switch type devices through the GUI.  Retain changes through the upgrade procedure and make it easy to change the Network Port configuration(s) through the node graphical interface instead of manually editing the swconfig file to support more than one DTD connection.

k1ky:
Although not directly related, how about migrating the swconfig file during upgrades? This would streamline our upgrades of MESH nodes that have customized DTD ports set up.

ae6xe:
The settings won’t be compatible on all devices after openwrt upgrade.  But maybe from this upgrade forward is an option.

cTurtle98:
just wanted to point out that this is similar to #466 and #588

aanon4:
With some more modern devices offering configuration (not via swconfig mind) this wont be implemented.

PR #834, opened 2023-05-11
Advanced networking tab

aanon4:
Proposing a new "Advanced Network" tab on supported devices (although I'm aware this is adding complexity). This provide UI access to features already available by editing configuration files. Two features are surfaced:

1. The ability to reassign the default purpose of the standard ports on supported multi-port devices.
2. The ability to create xlinks (vlans which act like vlan2) which we're using for wiring up backbone links in AREDN networks.

These features are currently supported by installing a seperate package, and we can continue to do that if people prefer, but I thought it might be a more generally useful feature to expose.

Visually, it looks like this:

Screenshot 2023-05-10 at 5 43 09 PM

Currently supported on the hAP ac2 and hAP ac3.

### DIFF
--- a/files/etc/httpd.conf
+++ b/files/etc/httpd.conf
@@ -6,4 +6,4 @@
 /cgi-bin/supporttool:root:$p$root
 /cgi-bin/advancedconfig:root:$p$root
 /cgi-bin/apiprotected:root:$p$root
-
+/cgi-bin/advancednetwork:root:$p$root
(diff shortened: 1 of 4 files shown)


dman776 (commented):
I like the concept.   Is a top-level menu the correct place for this?   Or should it be under Advanced Config somewhere?

aanon4:
We dont really have any other good UI spots for it. We have the "user" stuff (e.g. status page + other pages which dont require the admin password), the admin stuff (where I stuck this) and the advanced options page which is just a list of properties. You cant really put this UI there, and we dont have any call outs to new pages from that page - it wouldnt fit with the way that all works. So ... it ended up here as the best option.

dman776 (approved):

k1ky:
I believe this would satisfy my feature request #19?

aanon4:
@k1ky This is only for devices which support the new Distributed Switch Architecture (DSA) in OpenWRT I'm afraid.

PR #843, opened 2023-05-18
UI improvements from AB7PA

aanon4:


ab7pa (approved):
Tested on hAP ac2 and it looks great!  Nicely done, @aanon4

PR #858, opened 2023-06-01
Fix lua converting empty port object to empty array.

aanon4:
Lua confuses empty objects and empty arrays when converting to JSON.

Problem was observed by @ab7pa when editing port assignments.


ab7pa (approved):
This fixes the issue.  Thanks very much, @aanon4 Tim.

PR #860, opened 2023-06-03
More multi-radio support

aanon4:
Add support for multi-radio power settings.
Update various radio api to take wifi as argument rather than assuming a specific interface.


ae6xe (approved):
Looks good.

Issue #861, opened 2023-06-03
Code error reported in NB 20230603


Orv:
On a hAP ac 2:
Installed nightly build 202300603.
Unchecked 2 GHz Mesh, attempted to save and got:

/usr/bin/lua: /usr/lib/lua/aredn/hardware.lua:90: attempt to perform arithmetic on a nil value stack traceback: /usr/lib/lua/aredn/hardware.lua:90: in function '?' /usr/local/bin/node-setup:653: in main chunk [C]: ? 

Second save attempt yielded:

Bad Gateway
The process did not produce any response

Rechecked 2 GHz Mesh - save attempt yielded Bad Gateway again.

[supportdata-W6BI-SHACK-NODE-202306031816.gz](https://github.com/aredn/aredn/files/11643381/supportdata-W6BI-SHACK-NODE-202306031816.gz)

Orv:
NB 20230604-935dc2c resolves this issue.

PR #862, opened 2023-06-03
Handle non-wifi devices passed to maxpower/pwroffset functions

aanon4:
https://github.com/aredn/aredn/issues/861


ae6xe (approved):
looks good.

PR #865, opened 2023-06-05
Support xlinks on x86

aanon4:
x86 is a single port device, so the port section makes no sense for this device. But supporting xlinks is useful if you want to use x86 hardware (often quite a bit faster) for this feature.


mathisono:
some low profile boards do have 2ports. Dose the code based on available nic's?

aanon4:
The x86 is for virtualized hardware only as supporting lots of different x86 configurations is something we dont want to get into supporting.

PR #874, opened 2023-06-20
Merge all the station monitoring and mitigation into a single service.

aanon4:
This is an attempt to unify all the station monitoring and make it work better as one. We're trying to square a circle here somewhat, with taking steps to kick nodes when problems are detected, but not kick them too quickly or often in case we're mis-identifing issues.
We've seen these issue manifest themselves which nodes messing VoIP services as well as resets causing nodes to get into unrecoverable states when there was no real problems in the first place.
This will probably need to evolve before the next release, but would be good to get some milage on the new code.


aanon4:
There's also more stats recorded and the info included in the support tool data dump.

Issue #872, opened 2023-06-14
Feature Request: Group AAM messages


hickey:
I have been looking at the AAM feature and it is a handy way of sending out notices to the node owners / users of the AREDN network. But I think that there is one aspect to the AAM feature that should be explored and that is being able to send out "group" AAM messages. 

I see two primary use models for this feature. 
1. Being able to send a message to an administrative domain. This could be a geographical region, connection type (backbone vs leaf nodes vs tunnels), etc.
2. Message type distinctions: weather alerts, DHS threat level, wild fire dangers, infrastructure changes, NHL game results (just kidding on the last one :-)

There may be more applications than my initial thoughts here. As I am thinking about planning out a deployment of nodes,  I am trying to anticipate the operational needs of the network and coordinating across dozens of operators. I am just looking for ways to send out the alerts to nodes in a dynamic network where the core operators are not in control and may not really know the full network topology. 

The way that I envision this would work (and it seems pretty simple to implement) is that there would be another variable added to the advanced configuration for the AAM settings. This variable would contain a list of "labels" that get checked for messages in addition to the node name and the `all` text files. For example, if the variable contained the value `nfl` (North Florida) the polling of AAM messages would also look for `nfl.txt` and if found would display the message with any other AAM messages. Again, this would be a list so the variable could be a list that would look up each label in turn so node operators could subscribe to a number AAM messages.

PR #890, opened 2023-07-05
Added support for group alert messages

hickey:
These changes are to address the feature request that I put forward in #872. 

At this point the changes are untested, but they appear that they would work. If it is possible to build an AREDN firmware from this branch, then I would be very happy to test the changes on one of my nodes to validate that the functionality is working.

### DIFF
--- a/files/usr/local/bin/aredn_message.sh
+++ b/files/usr/local/bin/aredn_message.sh
@@ -33,6 +33,27 @@ true <<'LICENSE'
 
 LICENSE
 
+function retrieve_alert() {
+  url=$1
+  file=$2
+  label=$3
+
+  # label is an optional argument
+  [ -z "$label" ] && label=$file
+
+  wget -q -O "${file}" -P /tmp "${url}"
+  if [ -s "/tmp/${file}" ]; then
+    echo "<strong>&#8611; ${label}:</strong>"|cat - "/tmp/${file}" > /tmp/out &&
+    echo "<p>" >> /tmp/out &&
+    mv /tmp/out "/tmp/${file}"
+
+    # return success
+    return 0
+  else
+    return 1
+  fi
+}
+
 # does the node have access to downloads.arednmesh.org
 ping -q -W10 -c1 downloads.arednmesh.org > /dev/null &&
   online=true;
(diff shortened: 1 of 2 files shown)


ab7pa:
I have two suggestions.  First, make the _Alert Groups_ feature part of the "AREDN Alert Settings" **category** so they appear along with all of the other Alert Settings.  Second, the grouping feature is actually a subset of **Local Messages** so they should be under that code section rather than appending them separately at the end.  I'm attaching two files here:  one shows the Alert Groups as part of the _AREDN Alert Settings_ category, and the second shows what the entire **aredn_message.sh** script could look like with the groups as part of the _Local Message_ section of code.  I have tested both of these attached files on my Alert node and they work as you envisioned.

[advancedconfig-alertgroup.txt](https://github.com/aredn/aredn/files/11959955/advancedconfig-alertgroup.txt)

[aredn_message.sh.txt](https://github.com/aredn/aredn/files/11959959/aredn_message.sh.txt)

ab7pa:
Here is what the Alert Group code I suggested above looks like on one of my alert subscriber nodes:

hickey:
Ah. I was not looking at the fields close enough. The category field I was thinking that it was a field name. The intention was to keep the setting with the other alert fields. I will get the category field updated to get it included in the alert section. 

Your screen capture off one of your nodes is precisely what I was envisioning also. 

Are there any GitHub actions out there that I can add to my fork to build the sysimage file so that I can install and test changes to insure that things are working?  Actually with these changes the amount of updates are easy and I can go update one of my nodes (probably similar to what you did). It would be nice to have a way to generate the sysimage from a branch so that bigger changes can be tested easily.

hickey:
I have updated `advancedconfig` and  I am still working on `aredn_message.sh`. I have refactored out all the `wget` commands into a function since it is all the same operation and makes the code a bit more readable. I am looking at writing the final alert text to the file at same time as that would clean up a bit more code too. 

I have created https://github.com/aredn/documentation/pull/273 for documenting this feature.

hickey:
@ab7pa I have completed the updates to `aredn_message.sh`. Take a look and see if you see any issues. I have tested the local and group alerts and can not fore see any issues, but I am unable to test the AREDN messages. Ideally they should be handled in the same manner as before.

ab7pa (approved):
Looks good, @hickey   And thanks for the corresponding documentation update too.

aanon4 (approved):
Good to go.

PR #894, opened 2023-07-08
Corrected names of MikroTik of BaseBox models

hickey:
Please refer to https://www.arednmesh.org/content/mikrotik-basebox-2-unsupported

Because the current `radios.json` has the entry `Mikrotik RouterBOARD 912UAG-2HPnD` but the system info is reporting back `MikroTik RouterBOARD 912UAG-2HPnD` the unsupported banner is being triggered. This change fixes the case of the `T` in MikroTik so that the BaseBox series shows up correctly as a supported device.

### DIFF
--- a/files/etc/radios.json
+++ b/files/etc/radios.json
@@ -122,10 +122,10 @@
     "MikroTik RouterBOARD 952Ui-5ac2nD (hAP ac lite)" : {
       "maxpower"        : "22"
     },
-    "Mikrotik RouterBOARD 912UAG-2HPnD" : {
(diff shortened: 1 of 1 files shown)


aanon4:
Rather than chance the capitalization of what we have, could you add new entries with the new capitalization instead? Thanks.

aanon4:
Actually, scrub that. I think we'll finally bit the bullet and just lower case all the keys in the radios.json file.

aanon4:
https://github.com/aredn/aredn/pull/895

PR #895, opened 2023-07-08
Lowercase the keys in radios.json

aanon4:
Note. I also re-pretty-printed the JSON so it's more uniform.


ab7pa (approved):
Works well.  Thanks, @aanon4

PR #918, opened 2023-08-28
Simpler search

aanon4:
Unify the search fields into a single search at the top of the mesh page, rather than 9 search boxes for various fields. I think it's easier to find and use here.  Here's a visual. I also tweak the colors so it works when selecting different color styles.
Screenshot 2023-08-28 at 11 05 20 AM


ab7pa (approved):
Works perfectly and looks much cleaner.  Tested with all of the modified themes and all look fine.  Well done -- thanks @aanon4 Tim.

ab7pa (approved):
Tested latest commit and the Mesh Status page functions correctly.

Issue #919, opened 2023-08-30
[Solved] MikroTik hAP ac3 WLAN interface disappear / new ac3 comes in RouterOS v7


SCWhite:
### TL;DR
If you have new MikroTik hAP AC3
modify openwrt and build AREDN firmware from scratch

### So here is the story:
I got 3 of them, 1 old, 2 new.
the old one come with routerOS v6 (MAC DC:2C:6E:XX:XX:XX)
newer ones have routerOS v7 (MAC 48:A9:8A:XX:XX:XX)

I follow the document, and the old one works perfectly.
but no mater how I try to flash those new ac3, my wireless interface doesn't function at all.
(it seems to be broken on every nightly build / 3.23.4  / 3.23.8 )

And I keeps getting error message when I check "dmesg":

    ath10k_ahb a000000.wifi: failed to fetch board-2.bin or board.bin from ath10k/QCA4019/hw1.0

fortunately, someone on the forum give me the right direction
link: https://www.arednmesh.org/content/setup-hap-ac3-32340-problems

and I can verify it's the same issue I'm having.
link: https://forum.openwrt.org/t/no-wireless-mikrotik-rbd53ig-5hacd2hnd/157763

and here is that pull request from OpenWrt (still pending) :
https://github.com/john-tho/openwrt/pull/5
(I guess it have something to do with OpenWrt doesn't know how to unpack LZ77 data.)

so I pull AREDN from github, find the openwrt dir, add/modify the decompress method from @john-tho  and set config.mk to the right target.
(got a little hiccup when building, but it's not relevant)

finally, I have a working AREDN firmware for hAP ac3.

--
I want to thanks people on the forum / facebook group / github 
your advice really helps a lot.

Special shout-out to @john-tho , this won't be happening without his PR

so for people having latest version of MikroTik hAP AC3, there is a way to get your device working.

### And to save you some time, here is the firmware I built:

I really like @john-tho 's comment so I quote:
> usual disclaimer: expect your device to halt and catch fire…

[aredn-main-3635e22-ipq40xx-mikrotik-mikrotik_hap-ac3.zip](https://github.com/aredn/aredn/files/12474960/aredn-main-3635e22-ipq40xx-mikrotik-mikrotik_hap-ac3.zip)

SCWhite:
HI, @SensorsIot 
https://github.com/aredn/aredn/issues/813#issuecomment-1545493406
not sure if you're having the same issue with the new ac3 ?

[...]

PR #920, opened 2023-08-30
Add LZ77 decompression for newer Mikrotik devices

aanon4:
See conversation in this issue https://github.com/aredn/aredn/issues/919


SCWhite:
Thanks! I'm still struggling to learn how to write a patch.
You guy's are awesome!

aanon4:
While the OpenWRT systems makes senses, it not easy to use. In this case though, because you highlighted the original push request, it was easy to ask github to generate the patch for me.

PR #924, opened 2023-09-02
Add github templates for issues and PRs

BKasin:
When opening a new issue or pull request, templates will be provided that can be filled in.

In the case of Issues, 4 options will be given for a template to start with (bug report, feature/enhancement request, new hardware support, and general question), as well as an option for a blank issue. Choosing one of the template will automatically tag the issue. To see what this looks like, you can go here:

https://github.com/BKasin/aredn/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,24 @@
+---
+name: Bug Report
+about: Create a report for firmware related bugs.
+title: ''
+labels: 'bug'
+assignees: ''
+
+---
+
+### Describe the bug
+A clear and concise description of what the bug is.
+
+- [ ] Does this issue still occur on the latest version? (**Required**)
+
+### Expected behavior
+A clear and concise description of what you expected to happen.
+
+### Screenshots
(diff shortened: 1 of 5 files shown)


aanon4:
Would be good to get some feedback on this idea. On the one hand its a more structured way to report issues which could improve quality; on the other, forcing this structure on our users might be an unnecessary burden. Thoughts? @ab7pa @ae6xe @dman776 @K6AH

ae6xe:
I think it would help to obtain better quality information up front.   I don't think this will be much of a burden, as the requestor may readily ignore or not enter information in a give section, if they don't know what to do.

dman776:
I’m good with thisThanks,DarrylOn Sep 19, 2023, at 6:05 PM, Joe AE6XE ***@***.***> wrote:﻿
I think it would help to obtain better quality information up front.   I don't think this will be much of a burden, as the requestor may readily ignore or not enter information in a give section, if they don't know what to do.

aanon4:
I think people are good to merge this.

Unfortunately we're currently in a code freeze until the next release goes out, except for critical fixes. I will make sure we merge this once we release (about a month from now).

PR #930, opened 2023-09-16
Remember and reinstall packages after firmware upgrade

aanon4:
Remember what packages get installed by the user, and reinstall them on the next firmware upgrades.
Locally installed packages are stored and moved across with the upgrade.
It waits a couple of minutes after the upgrade to activate (so the network should have stablised).

Would be good to get some eyes on this to see how reliable it is.

### DIFF
--- a/files/etc/arednsysupgrade.conf
+++ b/files/etc/arednsysupgrade.conf
@@ -35,3 +35,4 @@
 /etc/local/uci/hsmmmesh
 /etc/passwd
 /etc/shadow
+/etc/package_store
--- a/files/etc/cron.boot/reinstall-packages
+++ b/files/etc/cron.boot/reinstall-packages
@@ -1,7 +1,8 @@
+#!/usr/bin/lua
 --[[
 
 	Part of AREDN -- Used for creating Amateur Radio Emergency Data Networks
-	Copyright (C) 2022 Tim Wilkinson
+	Copyright (C) 2023 Tim Wilkinson
 	See Contributors file for additional contributors
 
 	This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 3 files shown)


k1ky:
To be clear - this feature will begin to work for the next upgrade "after" N-20230916 is installed?

ab7pa:
Yes, the "package_store" will be created after you install 20230916 + your extra packages.  Once the package_store exists then it can be used by subsequent upgrades.

k1ky:
This may be the "coolest" upgrade feature since the inception of the ability to upgrade tunneled nodes remotely!  Great job!

k1ky:
Looks like there may be an issue with 32MB device upgrades for this feature.  NSM2 with IperfSpeed installed upgrading from Nightly 0916 to 0916 doesn't seem to bring the package over.  HAP and other 64MB devices seem to work fine.  Should I open an issue or note in documentation as modules aren't recommended or supported on 32MB devices for this feature?

aanon4:
There is no guarantee reloading packages will work as it depends on a lot of things, the network being stable at the moment this is happening being a major one. Much less likely to work on 32MB devices as they struggle.

And yes, if we dont already note this - and I thought we did on the admin page in red text - installing packages on 32M nodes is not a good idea.

k1ky:
Copy that.

ab7pa:
One great thing about any recent version of the firmware is that you do not need to use iPerfSpeed to test your throughput.  [The firmware already has the ability to do this](https://docs.arednmesh.org/en/latest/arednHow-toGuides/iperf.html#using-the-onboard-iperf-url-feature) without requiring any separate packages. (link to documentation explaining how to do this)

k1ky:
I really like the GUI features in the 0.8 Iperfspeed, being able to select available nodes without having to "type" everything in.

ab7pa:
The 32mb nodes and the docs both warn against installing extra stuff on low memory nodes.

ab7pa:
> I really like the GUI features in the 0.8 Iperfspeed, being able to select available nodes without having to "type" everything in.

The GUI comes at a cost. I find that browser bookmarks work nicely.

aanon4:
Some improvement: https://github.com/aredn/aredn/pull/932

PR #968, opened 2023-10-25
Wireguard tunnel support

aanon4:
Initial wireguard tunnel support.
Note: Needs to be merged after the OpenWRT 23.05.0 merge.


mathisono:
That was fast...

aanon4:
@ab7pa Yes, but you see much more right now because this PR is based on another pending PR, so you see all of that here too.

PR #975, opened 2023-11-09
Simplify how we measure link performance

aanon4:


ae6xe:
This looks like a lot more than the title suggests.   Is there a better title, or intent is to accept 38 commits?

aanon4:
This change is actually using the OpenWRT 23.05.0 merge as it's base (as it will go in on top of those changes), but this tool just show the diff against main, so you see all this other stuff too. Once the OpenWRT merge is made, this diff will become a single file change. This is the same for most other PRs pending right now.

I have no idea how to make this tool show something more sane. Maybe there's a way? You can generate the sensible diff yourself using the correct base.

PR #1003, opened 2023-12-09
Resolve download server ip by hand before download

aanon4:
Because we want to kill the dnsmasq server to save memory when downloading firmware, we should
resolve the download server IP by hand before killing it.
Switch to using curl so we can send the Host: header.


K1RKS:
Adding a comment for testing.

PR #1038, opened 2023-12-20
Simply the watchdog by fixing various timeouts.

aanon4:
Unfortunately there doesnt appear to be much flexibility in the various hardware watchdogs on radios, so setting the watchdog > 60 seconds mostly doesnt work. So rework the settings to allow for this and that our watchdog tests must be frequent and quick. It makes it less configurable and care needs to be taken when selecting ping targets (short timeouts)

### DIFF
--- a/files/usr/local/bin/manager.lua
+++ b/files/usr/local/bin/manager.lua
@@ -54,12 +54,12 @@ function wait_for_ticks(ticks)
 	local when = nixio.sysinfo().uptime + ticks
 	while true
 	do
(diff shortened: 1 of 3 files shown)


mathisono:
Could a watchdog be tied to the way nodes scan for NTP clock synchronization? Is a ping the the best way to determine if a node is operating correctly? What other normally according processes  can be tested? And would they return valid if a node stable in operation? Is the WLAN driver a simple thing to watch dog test without debug logging?

PR #1049, opened 2024-01-02
Make sure the node nodes we use for tunnels are always uppercase

aanon4:
Not sure how non-uppercase names make it into the /etc/config.mesh/vtun file, but make sure they are definitely uppercase when they're active.


k1ky:
I assume this applies to both Tunnel Clients as well as Tunnel Servers?

PR #1065, opened 2024-01-15
Added antennas to Ubiquiti NanoBridge M5

CoolAEW:
### Description
Added antennas to Ubiquiti NanoBridge M5 to get gain and beamwidth info

### Relevant Links
[Datasheet](https://dl.ubnt.com/nb5_datasheet)
[Antenna Info](https://dl.ubnt.com/UBNT-AOS_prod-specs-rf_3.pdf)

### Screenshots
None

### Additional context
The NanoBridge M5 comes in 2 versions with different dishes, but both use the same radio and will show as the same hardware.
Added antennas to combat this issue.

### DIFF
--- a/files/etc/antennas.json
+++ b/files/etc/antennas.json
@@ -323,6 +323,18 @@
             "beamwidth": 30
         },
         {
+            "model": "NBM5-22",
+            "description": "airMAX NanoBridge M5, 22 dBi 8&deg; dish;",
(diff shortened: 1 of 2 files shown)


aanon4 (commented):
Just want to confirm that this device is sold as a single radio, but with two dish options?

aanon4 (approved):

PR #1078, opened 2024-01-22
Deduplicate and sort tag names in Advanced DHCP Options selector.

pmilazzo:
### Description
Previously, multiple definitions for the same tag name resulted in duplicate entries in the tag selector for Advanced DHCP Options. Now, duplicates are removed and the tag names appear in alphabetical order in the tag selector.


aanon4:
Nothing wrong with this per say, but you could create the second table as you create the first, only inserting into the second if it's not already in the hash? Would save you a loop iteration.

pmilazzo:
If I understand you correctly, the second table is created from the first, so it is already deduplicated and there are no extra iterations. What I didn't do was realize that the table only needs to be built once and reused for each row. Now it is.

aanon4 (approved):

PR #1117, opened 2024-03-22
Improve speed of search

aanon4:
We were searching lots of empty fields, which just slowed things down for no reason.


aanon4:
And cache search values rather then re-extracting them from the HTML (which is slow)

PR #1135, opened 2024-03-29
Rewrite iperf to make it more reliable

aanon4:
Try to improve iperf reliability


aanon4:
Extensive rewrite to make iperf testing more resilient.

Issue #888, opened 2023-07-02
Ubiquiti AC devices dont return any devices during a scan


aanon4:
If you run a scan on a Ubiquiti AC device, you dont see any devices. In fact, the underlying command for this "iw wlan0 scan passive" doesnt return any devices except yourself (an active scan does the same). This works fine on non AC devices and AC devices from Mikrotik, so something low level is amiss here.

Orv:
Can we flag this as a bug?  That would be helpful when looking at the list of issues.

aanon4:
For my notes (and any one interested) here's a comparison of the WMI events (that basically the wifi chipset+firmware talking to the kernel) for an Mikrotik AC device compared to a UBNT AC device:

Mikrotik AC:
```
WMI_UPDATE_STATS_EVENTID
Thu Oct 19 23:40:00 2023 kern.debug kernel: [ 3265.460368] ath10k_ahb a800000.wifi: scan event foreign channel exit type 256 reason 6 freq 5640 req_id 40961 scan_id 40960 vdev_id 0 state running (2)
Thu Oct 19 23:40:00 2023 kern.debug kernel: [ 3265.461357] ath10k_ahb a800000.wifi: chan info err_code 0 freq 5640 cmd_flags 2 noise_floor -105 rx_clear_count -660641922 cycle_count -777560427
Thu Oct 19 23:40:00 2023 kern.debug kernel: [ 3265.461788] ath10k_ahb a800000.wifi: scan event bss channel type 4 reason 6 freq 5640 req_id 40961 scan_id 40960 vdev_id 0 state running (2)
Thu Oct 19 23:40:00 2023 kern.debug kernel: [ 3265.462185] ath10k_ahb a800000.wifi: chan info err_code 0 freq 5640 cmd_flags 1 noise_floor -105 rx_clear_count -660641922 cycle_count -777487445
Thu Oct 19 23:40:00 2023 kern.debug kernel: [ 3265.578259] ath10k_ahb a800000.wifi: scan event foreign channel type 8 reason 6 freq 5655 req_id 40961 scan_id 40960 vdev_id 0 state running (2)
Thu Oct 19 23:40:00 2023 kern.debug kernel: [ 3265.578820] ath10k_ahb a800000.wifi: chan info err_code 0 freq 5655 cmd_flags 0 noise_floor 0 rx_clear_count -658916258 cycle_count -762746815
[...]

PR #1144, opened 2024-04-01
Ubiquiti AC wifi scan fix

aanon4:
If we disconnect then reconnect the radio then the wifi scan on the Ubiquiti AC will work for a short while.


aanon4:
https://github.com/aredn/aredn/issues/888

PR #1156, opened 2024-04-07
Show progress bar during reboot/reflash and show estimated time remaining in MM:SS.

pmilazzo:
### Description
When rebooting or reflashing the node, display a progress bar as
well as the estimated time remaining. Also, show the time remaining
in MM:SS format.

### DIFF
--- a/files/usr/lib/lua/aredn/html.lua
+++ b/files/usr/lib/lua/aredn/html.lua
@@ -194,7 +194,7 @@ function html.wait_for_reboot(delay, countdown, address)
         const div = document.getElementById("countdown");
         if (div) {
             let t = Math.round(]] .. countdown .. [[ - (Date.now() - start) / 1000);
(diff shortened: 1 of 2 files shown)


aanon4:
Just confirming you've tested this patch before I commit it?

pmilazzo:
Tim: thanks for testing! I don't have a way to test with Safari, but in other browsers the Cylon effect means that the _value_ attribute of the `progress` element hasn't been set. I don't initialize _value_ in the original document, so the Cylon effect is expected until the first execution of the `cdown()` function. The Cylons will now also reappear if the estimated time expires, to suggest that the remaining interval is unknown (and perhaps that you should be relocating your Battlestar ASAP...).

I made a few tweaks I hope will improve browser compatibility; it seems to work with Chrome, Edge, and Firefox.

I have tested flashing both with and without retaining the configuration.

aanon4 (approved):
That seems to have made Safari happy.

PR #1160, opened 2024-04-10
Scan cache

aanon4:
Cache the last wifi scan and update it when a re-scan is requested. This change was suggested as a way of handling Ubiquiti AC devices which disconnect while scanning and making retrieving the results problematic if that was your connection. Now we scan and store the results so they can be retrieved later. In fact we no longer scan when navigating to this page show the last scan and a time stamp.  To update the scan push the rescan button. This make the page generally more responsive when initially navigated to.

### DIFF
--- a/files/www/cgi-bin/scan
+++ b/files/www/cgi-bin/scan
@@ -36,6 +36,7 @@
 --]]
 
 require("nixio")
+require("luci.jsonc")
 require("aredn.http")
 require("aredn.hardware")
 require("aredn.utils")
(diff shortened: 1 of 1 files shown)


mathisono:
I wonder if this could be problematic, and not show SSIDs all SSID's.  if
the broadcast interval of Wi-Fi SSIDs is missed.

aanon4:
@mathisono Can you explain more? This change doesnt effect how the scan actually works.

PR #1189, opened 2024-05-05
Fix github_changelog_generator name

aanon4:


k6ccc:
The link on the AREDN Nightly Builds page is now resulting in a 404 error.
https://www.arednmesh.org/content/nightly-builds
https://downloads.arednmesh.org/snapshots/CHANGELOG.md

Issue #1194, opened 2024-05-12
WorldMap Data Export Issue


Orv:
### Describe the bug
At least the .csv export from the WorldMap appears to exclude WireGuard tunnels.

- [ ] Does this issue still occur on the latest version? (**Required**)
Yes

### Expected behavior
All versions of tunnels should be included in all the data exports formats

aanon4:
Looks like an error in sysinfo.json where the node doesnt count these when generating the data.

Orv:
This works as expected.

PR #1195, opened 2024-05-12
Fix various status information errors

aanon4:
Fix various problems when reporting channel and frequency information on 3GHz and 900MHz nodes.
Fix tunnel reporting
Stop LQM copying mac file when it hasn't changed.


aanon4:
https://github.com/aredn/aredn/issues/1194

Issue #1208, opened 2024-05-16
Add antenna to device data


Orv:
### Describe the bug
The RocketDish RD-3G26 dish antenna is missing from the Rocket M3 antenna drop down.
- [X] Does this issue still occur on the latest version? (**Required**)
Yes
### Expected behavior
Should appear in dropdown

### Additional context
26 dBi gain, 7 degree beamwidth

ae6xe:
Many/All ubiquiti 'M' devices:  add the following antennas -  RD-2G24 RD-3G26 RD-5G30 RD-5G30-LW RD-5G34 AM-9M13 AM-2G15-120 AM-2G16-90 AM-3G18-120:

https://dl.ubnt.com/datasheets/airmaxsector/airMAX_Sector_Antennas_DS.pdf
https://dl.ubnt.com/datasheets/rocketdish/rd_ds_web.pdf

aanon4:
@ae6xe failure sure I have most of those - which ones am I missing?

aanon4:
If you folk could check that all listed above are either already in the antennas.json file or in this PR - thanks.
https://github.com/aredn/aredn/pull/1209

ae6xe:
On ~yesterday's nightly build, the Rocket M3 does not show any of the ubiquity antennas in the list.  The M5 does not show the non-ac airMax rocketdishes, the older models in use (different specs?).    The listings could be tightened up by showing the model #s of these antennas.   See attached pics.
![20240516_072650](https://github.com/aredn/aredn/assets/25602681/4813e92c-2370-4fa3-8912-1fcd94645370)
![20240516_072819](https://github.com/aredn/aredn/assets/25602681/308229f0-2167-4651-8672-b8d524ad6c36)

PR #1209, opened 2024-05-16
More antennas

aanon4:
https://github.com/aredn/aredn/issues/1208


Orv:
Needs these additional Ubiquiti omni antennas
[amo_ds_web.pdf](https://github.com/aredn/aredn/files/15366181/amo_ds_web.pdf)

Orv:
If not already present, add these Miktok antennas:

PR #1235, opened 2024-06-01
Fetch routes from kernel rather than OLSR

aanon4:
Fetch routes from kernel rather than OLSR
Fix xlink detection

### DIFF
--- a/files/etc/uci-defaults/80_aredn_lqm
+++ b/files/etc/uci-defaults/80_aredn_lqm
@@ -20,7 +20,7 @@ done
 
 if [ "$(/sbin/uci -q get aredn.@lqm[0].min_quality)" = "50" ]; then
     /sbin/uci -q set aredn.@lqm[0].min_quality=35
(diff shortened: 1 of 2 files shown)


Orv:
What's the benefit to this change?  (asking for a friend)  :-D

aanon4:
Fetching the routes from OLSR seems to make it crash - not always but sometimes. It happens less with fewer routes, and on faster nodes (never see it on my virtualize x86 nodes). As the same information can be taken from the kernel, I switch to doing that rather than asking OLSR. Hopefully this will avoid a few more crashes.

However the "mesh" page does this and it's unavoidable there, so at some point I'll need to work out why this can crash OLSR. I've looked but so far nothing jumps out. @ae6xe submitted the last fix to them about this problem and that clearly helped a lot; but the problem persists in some less frequent form still.

Orv (approved):
We'll give it a shot!

PR #1254, opened 2024-08-02
Add modules for GPS support

aanon4:
Add optional packages which can be installed to support USB GPS devices on nodes with USB.


ae6xe:
dusting off some cobwebs, there was also a serial related driver needed for precision time.  In my case, for the Rocket M5 GPS 0xe1c5 with gps chip off a serial port.  I forget the driver name...

aanon4:
This is enough for the USB GPS external devices. If you know what's missing for others then we should add those optional modules too.

ae6xe:
forum posts on this issue:  https://www.arednmesh.org/comment/21383#comment-21383

using i2c with the gps chip on the Rocket M5 gps, it's only going to be at best a stratum 1 timeserver.  If not using i2c, not sure what accuracy could be  claimed.  Just documenting this here in case others are going down this path.

PR #1255, opened 2024-08-03
update wg tun sv NET label

ab7pa:
### Description
Fix wireguard network field label

### Screenshots
Screenshot 2024-08-03 at 8 40 31 AM


aanon4 (approved):

K5RA:
The fields in the WG Tunnel Server and WG Tunnel Client windows have different names, even with this change.  
Suggest that Client side fields be named Server, Key, and Net.

PR #1257, opened 2024-08-04
First-use setup script to simplify initial setup of AREDN VMs.

aanon4:


aanon4:
A command line program to set the node's name and password immediately after the AREDN VM has booted. At this point it's address is 192.168.1.1 and it's not always easy to get to that in a virtualized environment. This makes the initial setup easy.

Issue #1269, opened 2024-08-16
Add ability to switch between 12 hour and 24 hour displays


Orv:
**Describe the solution you'd like**
Add ability to switch between 12 hour and 24 hour displays

aanon4:
Done

PR #1275, opened 2024-08-16
Support 24 hour clock display

aanon4:


aanon4:
https://github.com/aredn/aredn/issues/1269

Issue #1287, opened 2024-08-17
UBNT PowerBeam M5 300 not showing connected stations in new UI


k1ky:
### Describe e description of what the bug is.
RF and Tunnel Connected stations not showing on node list. N-20240817

- [x] Does this issue still occur on the latest version? (**Required**)
- [x] Yes

### Expected behavior
Should show stations.

### Screenshots
Please include a screenshot of your system information (if possible) if the specific system environment is relevant to the bug.
<img width="925" alt="image" src="https://github.com/user-attachments/assets/332642a2-5c60-4978-b885-8e9b24352ad6">
[supportdata-K1KY-PBELM5XW-149-MQY-2024-8-17-11-57.tar.gz](https://github.com/user-attachments/files/16645633/supportdata-K1KY-PBELM5XW-149-MQY-2024-8-17-11-57.tar.gz)

k1ky:
#1292 takes care of this on this model.

PR #1292, opened 2024-08-17
Handle old nodes with missing node details

aanon4:
https://github.com/aredn/aredn/issues/1287


k1ky:
Did this make it into N-20240818? If so, problem with PBELM5XW (PowerBeam M5 300 XW) still not showing connections.

aanon4:
I see connections on this node when I look at it.

PR #1356, opened 2024-08-22
More high contrast

aanon4:


K5RA:
I think your new High Contrast/Larger Font theme is major improvement over Light theme.

Issue #1344, opened 2024-08-21
Add WAN Client wifi status icon(s)


ab7pa:
Add the three wifi status icons to the **WAN Client** display (as appeared in the legacy UI).

![Screenshot 2024-08-21 at 7 52 14 AM](https://github.com/user-attachments/assets/2d19d6f1-e6a4-4737-8acf-78ba5a2f11ef)

aanon4:
A status word rather than the old icons.

PR #1362, opened 2024-08-22
Add Wifi Client status info

aanon4:
No icons - just words.


aanon4:
https://github.com/aredn/aredn/issues/1344

PR #1376, opened 2024-08-23
More high contast changes

aanon4:
Drop base font size from 19px to 18px.

### DIFF
--- a/files/app/resource/css/themes/high-contrast.css
+++ b/files/app/resource/css/themes/high-contrast.css
@@ -1,7 +1,7 @@
 :root
 {
     --font-family: sans-serif;
-    --font-base-size: 19px;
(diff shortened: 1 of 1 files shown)


K5RA:
I got 20240824 nightly this afternoon.  A Tunnel config Network:Port entry that has 3-2-3-3 IP adr cuts off last number of Port.  The IPs and MACs of DHCP LAN config are not 100%-all or half of last characters are obscured.   The only problem may be with the DHCP Reserved IPs.  Clicking on the cells brings up a drop-down list which is no help.   The Front Page still looks fine.   Whether this PR is left in or reverted is OK with me.  
I still owe you a study on DHCP LAN config with cursor pointing in each cell.

PR #1386, opened 2024-08-25
Remove functional use of :has css so we can support older browsers.

aanon4:
Some cosmetic use remains.

### DIFF
--- a/files/app/main/firstuse.ut
+++ b/files/app/main/firstuse.ut
@@ -72,13 +72,13 @@ if ((auth.isFirstUse || !config.authenable) && request.env.REQUEST_METHOD === "P
                     <div>
                         <small>This is the unique name given to your node. It must start with your callsign which must be capitalized. For example, <b>K6AH-home</b></small>
                     </div>
-                    <div class="cols">
+                    <div class="cols password-toggle">
                         <div>New Password</div>
-                        <div><input class="password-toggle" type="password" name="passwd1" required pattern="`[^#'"]+" minlength="4"><button class="icon eye"></div>
+                        <div><input type="password" name="passwd1" required pattern="`[^#'"]+" minlength="4"><button class="icon eye"></div>
                     </div>
-                    <div class="cols">
+                    <div class="cols password-toggle">
                         <div>Retype Password</div>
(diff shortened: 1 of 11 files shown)


jimsky7:
I am unfamiliar with when dailies are deployed. Installed 20240824-471c6a1 just now on http://aa6ax-twinpeaks-nsm5-ch177.local.mesh/a/status and the Login icon does not react to click. Later today I can look at the CSS. But maybe this isn't even the 'daily' that has the fix? I did the install using the node's old admin page. In Firefox I can 'inspect' the menu DIV and set 'display:block' at which point the Login menu appears, I can click and enter admin mode. So the only thing malfunctioning is the click on the icon. Hoping that I just don't yet have the version with the new code...

aanon4:
Unfortunately, due to GitHub failing at 2am this morning, the build is delayed and wont be available for another hour. You will need to refresh and redownload then.

jimsky7:
thx I understand this is not the build.

HOWEVER I (still) see an error that you'd want to address:

The FF debugger shows a 404 on this request:

http://localnode.local.mesh/a/css/admin.css.ac6e2dc517b471c95af9f00da2b92fc3

AND:

GET http://localnode.local.mesh/a/css/admin.css.ac6e2dc517b471c95af9f00da2b92fc3 net::ERR_ABORTED 404 (Not Found)

AND it claims there's a syntax error in this line "status:66":

        if (e.detail.requestConfig.triggeringEvent?.type === "click") {

in this block of code:

        if (e.detail.requestConfig.triggeringEvent?.type === "click") {
            const ismodal = e.detail.target?.id === "ctrl-modal";
            setTimeout(_ => {
                if (document.body.classList.contains("htmx-request")) {
                    logo.classList.add("animate");
                }
            }, ismodal ? 1000 : 500);
            if (ismodal) {
                setTimeout(_ => {
                    if (!e.detail.target.open) {
                        e.detail.target.showModal();
                    }
                }, 200);
            }
        }

No idea whether either of those is important or not, and also as you've said, my browser is a couple of years old. So Fyi.

aanon4:
The 404 are expected and are related to how resource loading is optimized.
The syntax error - sigh - I guess the browser's JS is also rather old.

aanon4:
https://github.com/aredn/aredn/pull/1393

PR #1393, opened 2024-08-25
Remove syntax not supported in older Javascript

aanon4:
?. syntax not available in older browsers.


jimsky7:
Login/Logout menus now appear and function. So far all admin functions that I've tried seem to work. They pop up a modal dialog and I am able to make changes and commit them. The 'commit' dialog functions.

PR #1403, opened 2024-08-26
 Fix parsing by new UI of advanced-DHCP-option configuration files.

pmilazzo:
### Description

- Fix spellings of suboptions and their keys.
- Do not drop leading spaces, so options with no tag will work.
- Do not drop trailing spaces, which may be required in tag patterns.
- Fix regular expressions to handle embedded spaces in options and tag patterns.
- Handle special characters in tag patterns (primarily for Subscriber-ID).


aanon4 (approved):
Thanks.

PR #1410, opened 2024-08-26
Focus newly created items

aanon4:
When we add new ports, services, xlink or tunnels, move the focus automatically to the first field of the new item.


pmilazzo:
This is awesome! Works great for:
- Add service
- Host aliases
- Port Forwarding
- Address Reservations (until you add an active lease)
- Add tunnel

Doesn't work for:
- Tags
- Options

PR #1429, opened 2024-08-30
Use autocomplete inputs to select nodes in tools

aanon4:
Apparently there's a HTML construct called datalists which are used with text input to provide autosuggestions. Since 2011. Who knew.


Orv:
:-)

PR #1440, opened 2024-09-01
Detect new UI in old server tunnels so we can warn

aanon4:
Because we can delete tunnels in the new UI, this will cause problems in the old UI which doesnt support this. Warn when we detect this.


Orv (approved):
Good catch!

PR #1471, opened 2024-09-07
Make index.html (which automatically redirects) respect dark mode

aanon4:


Orv (commented):
Would prefer "AREDN Node" over "Mesh Node"

aanon4:
That's what it's always said, but I can change it.

aanon4:
https://github.com/aredn/aredn/pull/1472

nc8q:
Would prefer "AREDN Node" over "Mesh Node"
+1 with Orv.

PR #1493, opened 2024-09-12
Remove race condition when counting base packages

aanon4:


K5RA:
I think this PR did not solve the problem.  I started a download of N 20240913, then went off to do other things, same back an hour later, and I am fairly sure the Packages count was zero until I clicked the area.  

System: hAP ac lite with latest nightly, R-Pi 5B updated yesterday to latest bookworm and Chromium 128.
Support data attached. 

[supportdata-K5RA-HAP-FB-2024-9-13-8-51.tar.gz](https://github.com/user-attachments/files/16995682/supportdata-K5RA-HAP-FB-2024-9-13-8-51.tar.gz)

K5RA:
Yes.  The package count was zero when node rebooted with new nightly this morning.  When I open/closed the Packages window, package count went to 1.

Out of curiosity, I downloaded N 20240913 on top of N 20240913 just now.  Installed packages count came up zero on reboot.  Refreshing the screen by clicking AREDN logo change it to one.  Easy workaround.

aanon4:
The package count is not live, so during an update it will be 0 until the background reload is done. And then you will need to refresh the browser to see the correct value. This is how its suppose to work.

PR #1523, opened 2024-09-17
Remove MAC from MAC/BSSID

aanon4:


ae6xe:
Hmmm.   When an AP sends a beacon it is the MAC address used as the BSSID -- copilot says:

when an Access Point (AP) sends a beacon in an 802.11 network, it includes the MAC address of the AP’s interface. [This MAC address is used as the Basic Service Set Identifier (BSSID), which uniquely identifies the AP within the network](https://en.wikipedia.org/wiki/Beacon_frame)

The MAC was there for other non-adhoc 802.11 connections where it is the MAC address of the radio's interface.

PR #1542, opened 2024-09-19
Add support for different wireless WAN client encryptions

aanon4:
Currently WAN client is always WPA2 or none. Provide WPA as an option.
It may make sense to add extra options here to handle random wifi networks people encounter.


aanon4:
Closing for now. Not clear there's a problem this needs to fix. Reopen if that changes.

PR #1559, opened 2024-09-22
Improve the initial nvram setup

aanon4:
Improve the initial setup on the nvram parameters, specifically mac2 and dtdmac. Avoid various race conditions when using the network devices to seed these values, and provide fallbacks if they fail.


aanon4:
More improvements in a new branch, so not using this one anymore.

PR #1560, opened 2024-09-22
Disable Done button if entries are invalid.

aanon4:
To help prevent the saving of incomplete (and there ignored) entries in multi-entry dialogs, disable the Done button when invalid content is detected. We do this for:
1. Tunnels
2. Services
3. Aliases
4. Forwards


aanon4:
I think there's a better way to do this - so wont commit this one.

Issue #1428, opened 2024-08-30
Allow a node's config to be backed up and restored


aanon4:
Provide a mechanism to save a node's current configuration and allow this to be restored at a later date.

Notes: it's probably useful to allow a config to be backed up from one type of and restored to another, even with different hardware. It wont be possible to guarantee it will work though and it's too hard to try to work out what combinations are okay. The same goes for firmware versions - might be okay to restore with a different firmware, but might not. Need some warning around this.

mathisono:
about time....

Thanks

PR #1597, opened 2024-10-03
Backup and Restore node configurations

aanon4:


aanon4:
https://github.com/aredn/aredn/issues/1428

files/app/main/status/e/backup-config.ut
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * You must not misrepresent the origin of the material contained within.
+ *
+ * Modified versions must be modified to attribute to the original source
+ * and be marked in reasonable ways as differentiate it from the original
+ * version
+ */
+%}
+{%
+if (request.env.REQUEST_METHOD === "PUT") {
```
rgeyer, line 36: I presume this method/page is only for downloading the backup file? If so, the only acceptable verb should be `GET`.

Also, strictly speaking, for a REST API, this would return a 405, with a body containing details on the acceptable verbs. Does the UI send a header or other token that could be used to distinguish?

```suggestion
if (request.env.REQUEST_METHOD !== "GET") {
```

files/app/main/status/e/backup-config.ut
```diff
@@ -0,0 +1,51 @@
...
+    response.headers["HX-Redirect"] = request.env.REQUEST_URI;
+    return;
+}
+response.override = true;
+const file = configuration.backup();
+if (!file) {
+    uhttpd.send(`Status: 404 Not Found\r\nCache-Control: no-store\r\n\r\n`);
```
rgeyer, line 43: Pedantic nit, but.. Is this the right HTTP code?

The server failed to find the file(s), which *should* have existed. A 404 definition states that the resource may be available in the future, and that subsequent requests are permissible. I don't think the availability of the backup files would change?

204 is No Content, which may be more applicable. The server processed the request, but had nothing to return, it also does not guarantee or imply that you could retry and have success.

I suppose this isn't actually a REST API, but.. 🤷

files/app/main/status/e/backup-config.ut
```diff
@@ -0,0 +1,51 @@
...
+}
+else {
+    const tm = localtime();
+    uhttpd.send(`Status: 200 OK\r\nContent-Type: application/octet-stream\r\nContent-Disposition: attachment; filename=backup-${configuration.getName()}-${tm.year}-${tm.mon}-${tm.mday}-${tm.hour}-${tm.min}.backup\r\nCache-Control: no-store\r\n\r\n`);
```
rgeyer, line 47: Is the user not meant to know that this is a `*.tgz` file? I presume there's no user readable data inside anyway, but I'm not convinced there is harm in allowing the user the hint on what the raw file type is.

files/usr/share/ucode/aredn/configuration.uc
```diff
@@ -428,3 +428,48 @@ export function unescapeString(s)
...
+    if (!fi) {
+        return null;
+    }
+    const fo = fs.open("/tmp/sysupgradefilelist", "w");
+    if (!fo) {
+        fi.close();
+        return null;
+    }
+    for (let l = fi.read("line"); length(l); l = fi.read("line")) {
+        if (!match(l, "^#") && !match(l, "^/etc/config/") && fs.access(trim(l))) {
```
rgeyer, line 446: More for my own edification, so you're welcome to ignore this but..

This implies that `/etc/arednsysupgrade.conf` contains a list of relevant config files for the node. It also implies that only some of the lines in that file refer to files in `/etc/config/`, and that only those files should be kept for backup purposes.

Why the extra steps here? Would it be okay to just compress the entirety of the `/etc/config` directory and send it down?

rgeyer (approved):
Dang, you beat me to it! I'll get my proxmox setup so that I can experiment and hopefully contribute in the future.

PR LGTM. No idea if my review will unblock this for merge, but I'm marking it approved.

I did add several pedantic suggestions/questions. Those are coming from the perspective of a user who might automate this externally. I.E. I have a Node-RED, task, or cron job, or kubernetes cron job, or some other automation that automatically downloads the backup(s) from one or more nodes.

That automation would be expecting more of a REST API type interface. That said, I suspect that most of the existing backend is expecting requests only from the UI, so maybe this is more of a "what if there were API compatibility" seed.

files/app/main/status/e/backup-config.ut
aanon4: It's not a REST API.
Yes this page is used to download the backup file, but it's not accessible via GET because we dont protect GETs from guest access.  Could I do that? - sure - but it's easier just to put something requiring admin access behind a PUT. This triggers a browser download and I'm not entirely sure what it'd do with a 405. But there's no application in the browser to interpret it anyway.

files/app/main/status/e/backup-config.ut
aanon4: The file is missing so I'm happy with a 404 and most people understand what it means. Not that it will ever happen anyway.

files/app/main/status/e/backup-config.ut
aanon4: I'm not trying to hide this from people, but I am trying to make the later upload option better at filtering what are acceptable backup files.

files/usr/share/ucode/aredn/configuration.uc
aanon4: AREDN doesnt use /etc/config like OpenWRT does. We dont backup any of /etc/config right now (if you look at the current release you'll see that's not a thing). However, we will start to include some very select bits of it in order to reduce the firmware upgrade time. But we dont need this and we dont want it in a backup, so it's excluded here.

PR #1605, opened 2024-10-06
Improve antenna settings messaging

aanon4:


K5RA:
I think Elevation should be between -90 and +90 degrees  from the horizontal.  I am not able to think of an antenna mount where this would be appropriate.

K5RA:
Excellent.

caseydiers:
@aanon4 
I also proposed reordering them so that azimuth and elevation were together:

- height
- azimuth
- elevation


 I find it to be more intuitive, than the current ordering of 

- azimuth
- height
- elevation

aanon4:
Is there a compelling case for any particular order? Perhaps how it's ordered in other software, or some generally accepted practice I can look at?

PR #1615, opened 2024-10-10
Wifisignal improvements

aanon4:


ab7pa (commented):
This fixes the reported issue.  Because you take the level back to zero, it will draw lines to zero on the graph periodically.

aanon4:
Just verifying you update both ends?

ab7pa:
Ah...  no, just the end I'm using for the graph.  Will update the other and report back.  THX

aanon4:
Can I reach the nodes you're testing?

ab7pa:
Installed the new wifisignal.ut on both ends, and the graphs look great now -- no zero-level values are plotted over the entire 5 min span.

PR #1630, opened 2024-10-14
VHT support (experimental)

aanon4:
* Enable VHT support on AC radios. This is experimental and I dont know if it will improve link performance in the real world.
* Switched to 'band' in wireless config as 'hwmode' was deprecated years ago.
* Fixed wifi scanner misreporting connected stations as APs.


ae6xe:
A benefit of supporting VHT is the higher 256-QAM MCS8 rate when the SNR/environment between radios allows it.    Should routinely work within the shack or shorter distances.  The table shows MCS9 needs 40MHz bandwidth with only 2 antennas -- not able to take advantage unless adding this BW setting to consider evaluating.    Hams in the less populated areas may be able to benefit.

PR #1632, opened 2024-10-15
Improve startup of LQM so we get some information early

aanon4:


mathisono:
Needed it takes to long, almost makes me think the node is broken.

PR #1648, opened 2024-10-25
Improve tunnel messaging

aanon4:


K5RA:
I think these are good improvements for WG Server and Client configuration pop-ups.   Maybe need to note in docs that "remote" is relative to the end you are configuring.  
Having key auto-fill with pop-up opening on the server side is good as well.

PR #1730, opened 2024-12-01
Fixes and tweaks for DHCP option in-browser validation

pmilazzo:
### Description
Fixes and tweaks for DHCP option in-browser validation:
- Many options take a list of IP addresses rather than a single address.
- Use <input type="number"> for values that are single integers.
- IP address pairs must be joined by commas, not whitespace.
- Add all the missing patterns.
- Support options that have no value.
- Minor pattern efficiency improvements.

Known issues:
- Styling has not been provided for the numeric input fields.
- Rapid-commit will not work because the --dhcp-rapid-commit option must also be provided.

### DIFF
--- a/files/app/main/status/e/dhcp.ut
+++ b/files/app/main/status/e/dhcp.ut
@@ -153,7 +153,7 @@ if (f) {
 f = fs.open(dhcp.dhcpoptions);
 if (f) {
     for (let l = f.read("line"); length(l); l = f.read("line")) {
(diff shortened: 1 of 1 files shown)


aanon4:
Number styling fixed here - https://github.com/aredn/aredn/pull/1731

aanon4:
While I see the 'rapidcommit' option documented in /etc/config/dhcp, I dont actually see it implemented in their config parser. I will add an implementation for it, but right now it wont actually do anything.

aanon4:
https://github.com/aredn/aredn/pull/1733

PR #1736, opened 2024-12-03
More fixes and improvements to DHCP option validation

pmilazzo:
### Description
More fixes and improvements to DHCP option validation:
- Handle unknown options (defined by number) as text data.
- Always reset to text input when changing to a non-integer data type.
- Tolerate values containing quotation marks.
- Make specifying "vi-encap:" encoding optional for Option 125 because it is the only supported encoding.
- Allow embedded quotation marks in string-encoded vi-encap data.

Known issues:
- Values containing embedded whitespace are not correctly encoded in /var/etc/dnsmasq.conf; each whitespace-separated token appears as a separate dhcp-option line. This problem is also present in both the old and new UI in previous releases.
- Nothing special is done with values that contain embedded quotation marks; these might confuse the dnsmasq option parser.

### DIFF
--- a/files/app/main/status/e/dhcp.ut
+++ b/files/app/main/status/e/dhcp.ut
@@ -158,7 +158,8 @@ if (f) {
     for (let l = f.read("line"); length(l); l = f.read("line")) {
         const m = match(replace(l, /\n+$/, ""), /^(\S*)\s(force|onrequest)\s(\d+)\s(.*)$/);
         if (m) {
-            push(advoptions, { name: m[1], always: m[2] === "force", type: int(m[3]), value: m[4] });
+            const p = replace(replace(replace(m[4], /"/g, "&quot;"), /</g, "&lt;"), />/g, "&gt;");
+            push(advoptions, { name: m[1], always: m[2] === "force", type: int(m[3]), value: p });
         }
     }
     f.close();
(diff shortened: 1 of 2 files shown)


convolved-KY4G:
Thanks for your efforts on this.  I've been testing this with nightly build 20241228-83e8462f some and decided to report out.  I've tried options 2, 3, 6, and 42 with a LAN-connected Grandstream GXP2160 phone and they appear to work.  The phone pulls the date/time if I allow access to the internet and specify an NTP server there.  I tried testing with whitespace per @pmilazzo above.  It looks as though the GUI will only accept IPs separated by a comma with no space (otherwise it won't accept the input).  I thought he was saying that the user could enter it, but it would be parsed wrong.  I tried using option 42 two IPs (only the second one being a true NTP server, and it worked that way as well.   I have not been able to get it to pull date/time info from the AREDN node running the whereandwhen app though.  I must have something configured wrong, so I'll have to keep investigating that.

convolved-KY4G:
Well, I got it working from the node itself.  You can see the details here [https://www.arednmesh.org/content/solved-grandstream-2160-will-not-get-correct-datetime-aredn-node-ntp-server#comment-24364](url).  If you know of a lighter weight way to do this, please respond in the linked mentioned above.

pmilazzo:
@convolved-KY4G Thank you very much for testing, and I'm sorry my description above was unclear.

Embedded whitespace is only accepted in option values that are encoded as strings or contain subparts encoded as strings. The dnsmasq server accepts option values encoded as "comma separated dotted-quad IPv4 addresses, []-wrapped IPv6 addresses, a decimal number, colon-separated hex digits and a text string." In the AREDN GUI, we restrict the encoding of values of known options to the most natural form for each option.

Option 42 ("ntp-server") uses the "comma separated dotted-quad IPv4 addresses" encoding and thus must not contain whitespace.

Currently, the following options are encoded as strings:
- 17 ("root-path")
- 18 ("extension-path")
- 40 ("nis-domain")
- 60 ("vendor-class")
- 64 ("nis+-domain")
- 67 ("bootfile-name")
- 77 ("user-class")
- 119 ("domain-search")
- any option not in the dnsmasq list of known options (use "dnsmasq --help dhcp" to get this list)

Additionally, the "option-data" portion of an encapsulated option within a [Vendor-Identifying Vendor-Specific Information Option](https://datatracker.ietf.org/doc/html/rfc3925#section-4) value can be specified as a string, as in:
`vi-encap:1234,56,"string value"`

If you would like more information about how dnsmasq handles option values, see the [DNSMASQ man page](https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html); search for:
`--dhcp-option=`

That said, embedded white space is, as I mentioned in my previous comment, not properly handled in the current version. We believe this problem will go away when AREDN merges [OpenWrt 24.10](https://openwrt.org/releases/24.10/start#openwrt_2410).

PR #1739, opened 2024-12-03
Attempt to stop the dhcp reservation wrapping.

aanon4:
Some people see this, but I cannot reproduce it, which makes fixing it tricky. So ... a guess.


aanon4:
@iannucci as I know you see this issue, could you check this?

PR #1740, opened 2024-12-03
Fix unresponsiveness when network changes

aanon4:
Tunnel changes, and possibly any network changes, which occur when we're holding a file system lock are making the entire node lockup. Not clear why this would happen, but disable the locking mechanism for now (was recently added as an extra layer of safety when making UI changes).


Orv:
I'll give it a shot.  Affects legacy tunnels only, or both types?

PR #1744, opened 2024-12-04
Basic syslog tool

aanon4:
A basic syslog tool for easy access to a node's logging.
There is only one log file now, so we dont need to access different ones.
No highlighting for now - what do people want to see? Are there good examples of this I can look at?

### DIFF
--- a/files/app/main/tools/e/syslog.ut
+++ b/files/app/main/tools/e/syslog.ut
@@ -32,16 +32,25 @@
  * version
  */
 %}
-<div>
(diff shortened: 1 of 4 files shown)


Orv:
A great question. Syslog somehow splits the logs it gets into over two dozen categorized files.  How does it do that?  What criteria?  We certainly don't need that level of highlighting.  I'll have to give that some thought.
On an unrelated issue - I need to learn how to use logrotate if I'm going to use syslog...

PR #1788, opened 2024-12-28
Move the LAN DHCP enable/disable option into the LAN DHCP panel

aanon4:
Separate the LAN DHCP server enable/disable from the network setup options, and move the switch to the LAN DHCP panel. You can have the LAN configured but the server disabled, which wasn't well represented in the UI before.


aanon4:
@ab7pa Just letting you know of this UI change.

convolved-KY4G:
Thanks for this change.  I had been hunting for where that setting was located.

Issue #1792, opened 2024-12-30
Stable release to Nighty upgrade (reset radio settings)


mathisono:
I have a LHG 5HPnD-XL running stable 3.24.10.
After I fetched the Update to nighty b18f.
The radio channel was reset to ch149/10 after the upgrade. This required direct access to the node to change.
I went with a hard line to the node, to gain access to the Node, I then changed the radio back to its previous channel and all is well now!

de KJ6DZB

aanon4:
We've seen this a few times. Any extra information you can remember about the node state before the upgrade would be appreciated because we dont know what causes it at the moment.

aanon4:
Fixed in tomorrow's build.

PR #1806, opened 2025-01-07
Fix channel 149 revert bug

aanon4:
The old configuration allowed the wifi interface to be blank. During an upgrade this caused the channel to reset to the default. The upgrade process now handles and fixes the blank interface and doesn't mess with the channel if already set.


aanon4:
Thanks to KJ6WEG for the support data which helped me finally track this down.

aanon4:
See also https://github.com/aredn/aredn/issues/1792

PR #1863, opened 2025-01-29
Further improvements to LAN to WAN firewalling.

aanon4:
Per comment in previous commit for this, add an explicity switch to control whether a LAN device can access the WAN across the Mesh.


aanon4:
So that is the question. Because it doesn't exist today it's currently "on". Off is a better default, but it will be a change people might not expect.

PR #1865, opened 2025-01-29
Allow local ntp server to run as a server for LAN devices

aanon4:


convolved-KY4G:
I tested the nightly build 20250129-d446fe15 (r28417-daef29c75d) on an AR300M16 with no external connectivity except for the GPS antenna.  The VOIP phone on the LAN picked up the date and time from the node ntp server as expected.  Thanks again for implementing this!

PR #1872, opened 2025-02-01
Older patch for Rocket 5AC Lite dts

aanon4:
There was an old rocket 5ac patch in previous kernels what help the network work at full speed. Experiment with bringing it into this kernel to see if it fixes what may be ethernet issue with some rocket 5ac lite hardware.

Experimental.


Orv:
Now works as expected.

PR #1899, opened 2025-02-17
Fix pasting into lat/lon fields

aanon4:


Orv:
I coulda swore that worked in the past...

aanon4:
It did - it got broken

Issue #1917, opened 2025-02-20
Unable to find AREDN Mesh Station from PtMp device


Orv:
### Describe the bug
Both devices on channel 149, 10 MHz BW

- [x] Does this issue still occur on the latest version? (**Required**)
Yes
### Expected behavior
PtMP station should see Mesh Station

### Support data

[supportdata-KB6DYJ-QTH-5G-2025-2-20-0-1.tar.gz](https://github.com/user-attachments/files/18879979/supportdata-KB6DYJ-QTH-5G-2025-2-20-0-1.tar.gz)
[supportdata-W6BI-TEST1-5G-2025-2-19-18-26.tar.gz](https://github.com/user-attachments/files/18879980/supportdata-W6BI-TEST1-5G-2025-2-19-18-26.tar.gz)

### Additional context
I think I saw it once, but unsure.

aanon4:
Can you try this experiment.

Log into KB6DYJ-QTH-5G
Type `/etc/init.d/wpad enable`
Reboot

aanon4:
I think this is the fixed, but the above should confirm it.
https://github.com/aredn/aredn/pull/1918

Orv:
[supportdata-K6PVR-CAM-HILLS-5G-2025-2-19-21-8.tar.gz](https://github.com/user-attachments/files/18881125/supportdata-K6PVR-CAM-HILLS-5G-2025-2-19-21-8.tar.gz)
[supportdata-W6BI-TEST1-5G-2025-2-19-21-10.tar.gz](https://github.com/user-attachments/files/18881124/supportdata-W6BI-TEST1-5G-2025-2-19-21-10.tar.gz)

A fresh set of support data files.

aanon4:
Unfortunately I dont capture the allowed mac address in the support data, but the station is being rejected by the PtP node, which suggests its wrong in some way.

k6ccc:
Home and connected to the LHG-2nD-XL.  Changed it to channel 1, rebooted, rebooted the hAP lite (not at the same time).  Refusing to connect.  Here's support data for both nodes.

[supportdata-K6CCC-RB-LHG-2nD-XL-2025-2-19-21-49.tar.gz](https://github.com/user-attachments/files/18881778/supportdata-K6CCC-RB-LHG-2nD-XL-2025-2-19-21-49.tar.gz)
[...]

PR #1918, opened 2025-02-20
PtxP fixes

aanon4:
Make sure wpad is running in station mode (not just ap mode)
Improve poor json parsing.


Orv:
Tried twice.  In this case, no change.

aanon4:
I need support data after you tried.

aanon4:
Put it on the bug https://github.com/aredn/aredn/issues/1917 and not here.

PR #1944, opened 2025-03-03
Enable zram swap (not on tiny)

aanon4:
If we have to swap, swap into RAM using zram.
Do we need swap?  There are lots of reports of the oom-killer shooting processes when there's plenty of memory available. There doesn't seem much consistency to the reporting, and various mitigation to disable oom-killer dont seem to disable it; at least when I try them. It has been noted in some Linux forums (Proxmox is one I'm familiar with) that Linux really wants to have swap and, until this patch, we have not provided it. It would make some sense that memory + swap is more tested than memory alone; and that Linux just really wants to swap sometime. So let's see if this helps.
ZRAM is the obvious choice here. Essentially it takes a percentage of memory and makes it into a compressed swap disk. This makes it fast, doesn't use of flash (which would be bad in many ways), and because it's compressed, kind of give us more memory to run apps; assuming our resident set size fits in "normal" memory.
For now this does not affect tiny devices. It is unclear how much RAM they could give to something like this.


aanon4:
Some other comments on this issue. It is claimed that when the linux kernel needs blocks of memory of a particular size, it will try to swap out pages to achieve this goal. But if it has no swap it cannot do this and may resort to killing processes to free up memory. It sounds reasonable, and also sounds like the kernel team would optimize for swap being available rather than implement other methods which would never get used. But, like much of this information, how much is guesswork and how much is known is unclear.

PR #1946, opened 2025-03-03
Lower min_free_kbytes to 1M (from default of 16M)

aanon4:
On low memory devices (<=32M) min_free_kbytes is set to 1M. On larger devices it is set to 16M. This is the memory the kernel keeps around for situations where it absolutely needs memory now or with crash. AREDN nodes are relatively stable memory wise, and this high limit is effectively reducing the memory we actually have to do work. So make it lower to improve stability.


aanon4:
4 hours of testing and things look stable. Usually fails by now.

PR #1989, opened 2025-03-15
Add babel mesh page navigation

aanon4:


ab7pa:
Instead of having a separate mesh view (slanted icon in left nav bar), would it be more intuitive to have a single mesh view which has a selectable filter (dropdown or radio button) at the top to display all nodes vs. only babel-enabled nodes?

PR #2017, opened 2025-03-23
LQM monitoring simplifications

aanon4:


aanon4:
Admittedly, not looking simpler ... but on the path to that one day.

PR #2024, opened 2025-03-26
Enable ed25519 ssh keys

myzhang1029:
`ed25519` ssh keys are ubiquitous and has become the default as of OpenSSH 8.5. I believe it would be nice for aredn to support it.


aanon4:
Thanks for this. I'll build and put it on a test node tomorrow. What the easiest way for me to test it?

myzhang1029:
My test was to generate a `ed25519` key (`ssh-keygen -t ed25519`), then upload the `id_ed25519.pub` file on the Web interface at Basics->Upload SSH Key.

Finally, I ssh'd into the router. If the update works correctly, it should not ask for a password anymore.

The pub key should also appear at the end of `/etc/dropbear/authorized_keys` on the router.

aanon4:
Thanks. One problem I see is that because the tiny build wont have this change (thanks for that by the way), the code change in basics.ut is incorrect for tiny builds. There needs to be a way for tiny builds to not accept these kinds of keys. Perhaps there's a way for dropbear to provide information on what key types it supports? Or some other way?

myzhang1029:
I dd not actually have access to a ath79-tiny device to see if it would cause any problem. I added the last commit because the tiny builds were supposed to "save as much space as we can", and Ed25519 is definitely something optional. What do you think @aanon4?

myzhang1029:
> Perhaps there's a way for dropbear to provide information on what key types it supports? Or some other way?

As for this, it seems that dropbear does not have something like OpenSSH's `ssh -Q key`. If we have to check for ed25519 support at runtime, it looks like we can do `dropbear --help` and `grep` for `ed25519`. Alternatively, is there a way to access configuration variables inside `basics.ut` at built time?

aanon4:
I'm assuming (I've not have time to build your changes yet - apologies) that a new host key is generated in /etc/dropbear (something like dropbear_ed25519_host_key maybe?). If so you could use the presence of this key to decide whether to accept the key in basics.ut?

Also, assuming this key does get generated, you need to add it to the backup system so it will be saved between updates.

myzhang1029:
> Also, assuming this key does get generated, you need to add it to the backup system so it will be saved between updates.

You are absolutely right. I was so confused why I'm getting a changed host key error today. I will check this and try to make a patch later today.

myzhang1029:
I believe Tim is referring to the size increase to the dropbear binary for including ECC algorithms. I will make two ath79-tiny builds later today to see if there is an appreciable difference.

myzhang1029:
(this comment is a WIP. Will be appending as builds finish)

Format is (bytes and filename) (output of `find openwrt/build_dir/target-<TARGET>/dropbear-2024.86/ipkg-<TARGET> -type f -exec wc -c {} \; | sort -h`)

With Ed25519 (myzhang1029/aredn@451bfaf5):

- Target `ath79-tiny`:

- Target `ath79-generic`:
```
21 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/CONTROL/conffiles
117 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/CONTROL/prerm
144 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/lib/upgrade/keep.d/dropbear
160 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/CONTROL/postinst
221 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/etc/config/dropbear
556 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/etc/uci-defaults/50-dropbear
580 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/CONTROL/control
992 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/lib/preinit/99_10_failsafe_dropbear
12314 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/etc/init.d/dropbear
197861 openwrt/build_dir/target-mips_24kc_musl/dropbear-2024.86/ipkg-mips_24kc/dropbear/usr/sbin/dropbear
```
Final `sysupgrade.bin` (taking `glinet_gl-ar150` as example)
```
6226759 openwrt/build_dir/target-mips_24kc_musl/linux-ath79_generic/tmp/aredn-nocall-ed25519-testbuilds-77bead1e-ath79-generic-glinet_gl-ar150-squashfs-sysupgrade.bin
```

- Target `ipq40xx-mikrotik`:

Without Ed25519 (myzhang1029/aredn@d4f0498d):

- Target `ath79-tiny`:

- Target `ath79-generic`:

- Target `ipq40xx-mikrotik`:

files/app/main/status/e/basics.ut
```diff
@@ -242,7 +249,7 @@ if (request.env.REQUEST_METHOD === "DELETE") {
             <div class="cols">
                 <div>
                     <div class="o">Upload SSH Key</div>
-                    <div class="m">Add SSH key</div>
+                    <div class="m">Add SSH key {% if (!detectEd25519()) { %} (no Ed25519 support) {% } %}</div>
```
aanon4, line 252: Let's not do that here. If you look a little lower in the file you'll find some inline help. If you want to add something there there about what keys are supported in both cases I would prefer that. I think telling someone what is supported rather than telling them what isn't would be better messaging.

myzhang1029: That is a good idea.

myzhang1029:
@aanon4 Would you be fine with me also adding support for `ecdsa-sha2-nistp384` and `p521` variants (the firmware already allows them and we just need to fix the RE)?

aanon4:
Good stuff.
Last minor thing I think - add yourself to the bottom of the CONTRIBUTORS files.

myzhang1029:
Thank you!

aanon4 (approved):

PR #2030, opened 2025-03-27
Stop upstreaming .local.mesh to DNS server

VA2XJM:
adding the `local` value will stop DNSMASQ to send the mesh specific DNS resolution to upstream servers. This considerably reduce the number of requests made to upstream server and also reduce resolution time when upstream server is not responding.

### DIFF
--- a/files/etc/dnsmasq.conf
+++ b/files/etc/dnsmasq.conf
@@ -8,6 +8,10 @@ expand-hosts
 # Never send any AAAA responses
 filter-AAAA
 
+# Domain + local (will not submit .local.mesh upstream) 
+domain=local.mesh
+local=/local.mesh/
+
 # Don't cache the errors
 no-negcache
 


files/etc/dnsmasq.conf
```diff
@@ -8,6 +8,10 @@ expand-hosts
 # Never send any AAAA responses
 filter-AAAA
 
+# Domain + local (will not submit .local.mesh upstream) 
+domain=local.mesh
```
aanon4, line 12: You dont need this. This is defined in the auto-generated dnsmasq and comes from values defined in the UCI state. It was recently removed because it was redundant.

files/etc/dnsmasq.conf
```diff
@@ -8,6 +8,10 @@ expand-hosts
...
+local=/local.mesh/
```
aanon4, line 13: Unless I misunderstand how this works, this would prevent the server from querying the supernode for non-local  lookups. In fact the supernode config contains an explicit directive for this. For example (from my localnode):

`server=/local.mesh/10.80.220.23`

aanon4 (commented):
I dont think this will work.

files/etc/dnsmasq.conf
VA2XJM: My understanding is:

The domain value will add `.local.mesh` at the end of a string not containing a dot for example if you try to resolve a hostname without domain, `someone-node`, it will use and return `someone-node.local.mesh` for the resolving.
> domain=<domain>[[,address range[,local]]|interface]
>    Specifies DNS domains for the DHCP server. Domains may be be given unconditionally (without the IP range) or for limited IP ranges. This has two effects; firstly it causes the DHCP server to return the domain to any hosts which request it, and secondly it sets the domain which it is legal for DHCP-configured hosts to claim. The intention is to constrain hostnames so that an untrusted host on the LAN cannot advertise its name via DHCP as e.g. "microsoft.com" and capture traffic not meant for it. If no domain suffix is specified, then any DHCP hostname with a domain part (ie with a period) will be disallowed and logged. If suffix is specified, then hostnames with a domain part are allowed, provided the domain part matches the suffix. **In addition, when a suffix is set then hostnames without a domain part have the suffix added as an optional domain part**. Eg on my network I can set --domain=thekelleys.org.uk and have a machine whose DHCP hostname is "laptop". The IP address for that machine is available from dnsmasq both as "laptop" and "laptop.thekelleys.org.uk". If the domain is given as "#" then the domain is read from the first "search" directive in /etc/resolv.conf (or equivalent).

The `local` value will ensure that the any request containing `.local.mesh` will not be passed to upstream server UNLESS we have a `server` line specified which take precedence.
> --local, --server=[/[domain]/[domain/]][server[#<port>]][@interface][@source-ip[#port]]
>    Specify upstream servers directly. Setting this flag does not suppress reading of /etc/resolv.conf, use --no-resolv to do that. If one or more optional domains are given, that server is used only for those domains and they are queried only using the specified server. This is intended for private nameservers: if you have a nameserver on your network which deals with names of the form xxx.internal.thekelleys.org.uk at 192.168.1.1 then giving the flag --server=/internal.thekelleys.org.uk/192.168.1.1 will send all queries for internal machines to that nameserver, everything else will go to the servers in /etc/resolv.conf. DNSSEC validation is turned off for such private nameservers, UNLESS a --trust-anchor is specified for the domain in question. An empty domain specification, // has the special meaning of "unqualified names only" ie names without any dots in them. A non-standard port may be specified as part of the IP address using a # character. More than one --server flag is allowed, with repeated domain or ipaddr parts as required.
>
>    More specific domains take precedence over less specific domains, so: --server=/google.com/1.2.3.4 --server=/[www.google.com](http://www.google.com/)/2.3.4.5 will send queries for google.com and gmail.google.com to 1.2.3.4, but [www.google.com](http://www.google.com/) will go to 2.3.4.5

So at this point, we stop passing DNS resolution requests to external (Google by default) DNS, wich reduce network usage (and `leaking` information?). But we send all request to supernode since the `server` value of `local.mesh` toward a supernode server is more specific than `local` alone. We may have a problem here if supernode is far away, slow to reply or is not aware of a certain set of host. _Should there be a new domain to use for supernode ? like remote.mesh ? no idea of the implication needed for that, just a quick thought_.

Also the fact that `domain-needed` is specified, it will try to resolve locally `node-name` but wont pass it to upstream server unless it contains the domain `.local.mesh`. So without a domain specified, we may have more troubles.

> domain-needed
>    Tells dnsmasq to never forward A or AAAA queries for plain names, without dots or domain parts, to upstream nameservers. If the name is not known from /etc/hosts or DHCP then a "not found" answer is returned.

VA2XJM: I can confirm that we above config, DNS resolution for "local mesh" is made using `addn-hosts=` and `hostsdir=` prior to be send to Super Node when the hostname is unknown locally. 

External DNS resolution requests are forwarded toward configured DNS server as supposed since their domain is not considered local.

This greatly reduced our problem, but not eliminated everything. I identified the most problematic systems we have generating alerts in the monitoring system and we disabled IPv6 on those server to see if this is a game changer or not. Results to come later.

aanon4: So I think what you're saying is that `local=/local.mesh/` will prevent this domain using the generic upstream resolvers but will not prevent the use of the supernode specific `server=/local.mesh/...` resolver. Do I have that correct?

aanon4: It is not clear to me still that, if you have the supernode `server=/local.mesh/...` setting, that dnsmasq would use the generic upstreams anyway.

VA2XJM: It will first try to resolve locally with hosts file, OLSR and Babel , if no answer it will upstream to specified `server=` values depending of other filters sets.

For requests outside of the mesh identified with a different domain, the configured upstream will be used (dafault to Google).

Here are an example:

```
# Device local to the node
Thu Mar 27 12:24:01 2025 daemon.info dnsmasq[11689]: query[A] ars-02.local.mesh from 10.170.72.227
Thu Mar 27 12:24:01 2025 daemon.info dnsmasq[11689]: /etc/hosts ars-02.local.mesh is 10.170.72.226

Thu Mar 27 12:24:05 2025 daemon.info dnsmasq[11689]: query[A] ve2atu-pbx02 from 10.170.72.227
Thu Mar 27 12:24:05 2025 daemon.info dnsmasq[11689]: /etc/hosts ve2atu-pbx02 is 10.170.72.227


# Device inside the mesh but attached to another node
Thu Mar 27 12:24:05 2025 daemon.info dnsmasq[11689]: query[A] ve2atu-zabbix.local.mesh from 10.170.72.226
Thu Mar 27 12:24:05 2025 daemon.info dnsmasq[11689]: /var/run/hosts_olsr ve2atu-zabbix.local.mesh is 10.87.43.77

# Device offline at the moment (same results with device outside of the mesh, so forwarded to the supernode (ip: 10.35.234.234))
Thu Mar 27 12:24:07 2025 daemon.info dnsmasq[11689]: query[A] ve2rui-fn08lf-nsm180.local.mesh from 10.170.72.226
Thu Mar 27 12:24:07 2025 daemon.info dnsmasq[11689]: forwarded ve2rui-fn08lf-nsm180.local.mesh to 10.35.234.234
Thu Mar 27 12:24:07 2025 daemon.info dnsmasq[11689]: query[AAAA] ve2rui-fn08lf-nsm180.local.mesh from 10.170.72.226
Thu Mar 27 12:24:07 2025 daemon.info dnsmasq[11689]: forwarded ve2rui-fn08lf-nsm180.local.mesh to 10.35.234.234

# Remode server via Internet
Thu Mar 27 12:28:04 2025 daemon.info dnsmasq[11689]: query[A] toronto7.voip.ms from 10.170.72.227
Thu Mar 27 12:28:04 2025 daemon.info dnsmasq[11689]: forwarded toronto7.voip.ms to 8.8.8.8
Thu Mar 27 12:28:04 2025 daemon.info dnsmasq[11689]: reply toronto7.voip.ms is 208.100.60.56
Thu Mar 27 12:28:04 2025 daemon.info dnsmasq[11689]: reply toronto7.voip.ms is <RRSIG>
```

aanon4:
Based on this discussion I think this is probably the approach to take: https://github.com/aredn/aredn/pull/2033

PR #2033, opened 2025-03-27
Improve supernode detector for when there isn't a supernode.

aanon4:
When a supernode had been detected we setup DNS forwarding for local.mesh so we could find unknown names. Now if the supernode isn't detected we setup a null forward (ie. there isn't one) so we dont send these lookups to the standard upstream DNS server which wont know anything about local.mesh


aanon4:
Thanks for the input from @VA2XJM on this.

PR #2116, opened 2025-05-11
add jq package option

ab7pa:
What would you think about adding the option to install the jq package?


aanon4:
What is it?

aanon4:
Also, I dont think this is sufficient to include it in the build as it's not part of the defaults. You also need to install it using the feeds command. Look in the top level Makefile for examples.

ab7pa:
After more thought, features of jq are already available in ucode, so the use of jq would be an inefficient way to accomplish json parsing tasks. All such tasks should be done using ucode instead.

PR #2157, opened 2025-05-20
Enable channels 180-184 in PtXP modes

aanon4:


AI7NC:
So glad we were able to figure this out, we were already starting to consider how we'd rework our band plan down here to accommodate the limitation for some connections.

PR #2165, opened 2025-05-22
Completely rewrite the routing rules.

aanon4:
This had got messy, difficult to understand, and in some cases just broken.


aanon4:
Going to require a fair bit of testing before this is committed as it could break nodes extremely badly if its wrong. Anyone who wants to look over this please do.

K5RA:
Looks like typo in new line 22 of Routing tables and their uses"  "thought" should be "through"?

In new line 44, maybe "which is" should be "which are" if refers to "rules"

PR #2255, opened 2025-06-18
Disable wpa_supplicant if we're not using encryption even if we need hostapd

aanon4:


aanon4:
This saves a bunch of memory and doesnt appear to be necessary unless you're using encryption in LAN or WAN mode.

PR #2363, opened 2025-07-25
Add two more packages from feeds to make openwrt build stop complaining along with fixes for mikrotik and tplink units

grozzie2:
The openwrt build complains these packages are missing.
Trivial change to the makefile to add them, just makes it look cleaner while building.

Worked on this more later, added some changes for TpLink CPE710.  NOHT is only needed for adhoc, not for sta and ap modes.  Moved qca9888 detection into the stanza for setting adhoc mode,  which leaves HT20 for sta and ap modes.  Makes a big difference in performance if using those modes.

Tested here in a 5 node mesh, 2 tplink, 2 ubiquiti and one mikrotik.

Added another set of changes, this time for the Mikrotik LHG 5 ac.  It suffers the exact same issue as the tplink unit did.  This change forces NOHT on the ipq4019 chipset, similar to the way it was done for the qca9888 on the tplink.

A happy side effect of this change, 5mhz channels work correctly now as well.

These changes are all extensively tested in adhoc, ap and sta modesm each with 5, 10 and 20mhz channel widths.  I am testing on a fully interconnected mesh that consists of 2 tplink cpe710, 1 ubiquiti litebeam 5 ac gen2, 1 ubiquiti rocket ac and one mikrotik lhg 5 ac xl.  Network is stressed by running multiple iperf runs across all link.

With NOHT forced for the qca9888 and ipq4019 in adhoc mode, the ath10k driver no longer abends when under load.

These fixes are a band-aid to deal with a driver issue.  Will see if we can fix the dirver in the near future and get the performance of HT20 back.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,8 @@ feeds-update: stamp-clean-feeds-updated .stamp-feeds-updated
 	cd $(OPENWRT_DIR); ./scripts/feeds uninstall -a
 	cd $(OPENWRT_DIR); ./scripts/feeds update -a
 	cd $(OPENWRT_DIR); ./scripts/feeds install libpam
+	cd $(OPENWRT_DIR); ./scripts/feeds install libev
+	cd $(OPENWRT_DIR); ./scripts/feeds install libtirpc
 	cd $(OPENWRT_DIR); ./scripts/feeds install libcap
 	cd $(OPENWRT_DIR); ./scripts/feeds install jansson
 	cd $(OPENWRT_DIR); ./scripts/feeds install libidn2
--- a/files/etc/radios.json
+++ b/files/etc/radios.json
@@ -576,7 +576,6 @@
   },
   "mikrotik lhg 5 ac (rblhgg-5acd)": {
     "maxpower": 25,
-    "exclude_bandwidths": [ 5 ],
     "antenna": {
       "description": "24.5 dBi 7&deg; Dish",
       "gain": 24.5,
@@ -585,7 +584,6 @@
   },
   "mikrotik lhg 5 ac xl (rblhgg-5acd-xl)": {
     "maxpower": 25,
-    "exclude_bandwidths": [ 5 ],
     "antenna": {
       "description": "27 dBi 7&deg; Dish",
       "gain": 27,
--- a/files/usr/local/bin/mgr/wireless_monitor.lua
+++ b/files/usr/local/bin/mgr/wireless_monitor.lua
@@ -93,7 +93,11 @@ function M.reset_network(mode)
         elseif mikrotik_ac then
             -- Only observered on Mikrotik AC devices
             os.execute(IW .. " " .. wifi .. " ibss leave > /dev/null 2>&1")
-            os.execute(IW .. " " .. wifi .. " ibss join " .. ssid .. " " .. frequency .. " HT20 fixed-freq > /dev/null 2>&1")
+            if nixio.fs.stat("/lib/firmware/ath10k/QCA4019") then
+              os.execute(IW .. " " .. wifi .. " ibss join " .. ssid .. " " .. frequency .. " NOHT fixed-freq > /dev/null 2>&1")
+            else
+              os.execute(IW .. " " .. wifi .. " ibss join " .. ssid .. " " .. frequency .. " HT20 fixed-freq > /dev/null 2>&1")
+            end
         else
             nixio.syslog("notice", "-- ignoring (mikrotik ac only)")
         end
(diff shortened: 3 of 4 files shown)


grozzie2:
When I was doing them, every time I tried to make a new pull for a separate commit, github just added them to the existing one.

I'm leaving on holiday tomorrow to take grandson camping, will re-vist this when we get back.

aanon4:
Closing. Please split any changes into PRs for each specific piece.

aanon4:
Following up on this again. Please note that development for these changes has moved into the Babel only branch. Right now the bulk of the work is in hardware.uc:getHTMode which attempts to make a better HT selection based on various factors. I suspect it is unduly pessimistic so thoughts appreciated.

aanon4:
> Going back to my notes from when I had a rigorous test environment set up.
> 
> In PTP and PTMP modes, setting HT20 on the CPE710 works just fine. I did find some interesting anomolies around bandwidth tests. All my tests were done running multiple iperf3 runs, everything fully automated via scripts. What was a bit puzzling, the thruput when set to 10 mhz bandwidth is identical to what it gets when set to 20mhz bandwidth. I found that puzzling, and it's something I want to revisit where there is time available.

1. Do you suspect that 10 mhz isnt being honored in PTP/PTMP modes? If so I would expect a laptop wifi scan would pick up the SID; something it doesnt do for 10MHz adhoc but does at 20 MHz adhoc. A better test would require a signal analyzer.
2. If you ran these tests with NOHT mode, did you see the same iperf for 10 and 20 MHz? Just wondering if the HT20 is overriding the 10 MHz setting in this scenario.

> 
> Adhoc mode is a very different beast. The original test setup consisted of two CPE710, one ubiquiti rocket ac, and one ubiquiti litebeam ac, all brand new at the time. Setting HT20 in all units, iperf thruput between the litebeam and rocket was as expected. connections to the CPE710 would start to run, then fail. Radios disconnected, and a full restart of the radio required to get it going again.
> 
> I ran a lot of tests, lots of different configuratios to narrow down the problem, ended up with this result.
> 
> Set the two CPE710 to NOHT, then reboot. All of the connections worked after that, and worked reliably. The two ubiquiti units got expected thruputs for an HT20 connection on a strong signal. Links between the two CPE got somewhat lower, and essentialy the same numbers between the cpe and the ubiquiti nodes. But there is one more important detail.
> 
> If I had the cpe unit running in HT20, saw it die, then just restarted wireless with NOHT, it would appear to run well, 3 or 4 iperf runs would go thru, but after a few more, it would die again. If I did a full reboot of the unit after failure and changing it to use NOHT, it would run reliably, scripted iperf session would continue for an hour without hiccup. turns out, it's important to never set HT20 on those units if in adhoc mode, because the failure that happens is not cleared up completely with just a reset of the radio, to clear it completely requres a full restart. I didn't go so far as to try unloading modules and reloading them to see if that was 'good enough' as the proper solution is to never set HT20 on the unit and have it try make an HT connection in adhoc mode.
> 
> The NOHT is only required for adhoc modes on the CPE710.

For my clarification. NOHT mode is required for adhoc mode on the CPE710 regardless of the bandwidth setting? Even 20MHz?

> 
> I got the Mikrotik unit with the ipq4019 chipset somewhat later, and the first time I saw the error in the kernel logs, it was the exact same error as the cpe710 gave, an abend in the ath10k driver from a null pointer. I ran the same tests, and got essentially the same results as the cpe710, so I changed the aredn stuff I have here to do the same, NEVER set HT20 into the radio if adhoc mode is being used. It ran stable, so the I tried something else, re-enabled 5mhz channels for the mikrotik, and voila, they worked too. What I did find in the early testing, 5mhz channels cause the kernel abend much quicker than 10 or 20, and often it happened just with a couple pings meant to flesh out arp tables etc when bringing up the adhoc test environment I have,

So NOHT for adhoc modes on the 4019 chipset (Mikrotik ac2, ac3 & sxt ac are the radios effected I think - Mikrotik also do some ac radios with a different chipset). Again. I want to clarify, that this is true for 5, 10 and 20 MHz channels?

Did you happen to touch VHT20 mode on any of these radios, specifically when they were at 20 MHz?

grozzie2:
yes, with adhoc, NOHT is required, irrelavent of bandwidth.  And yes, I did VHT.  Set uci to VHT, iwinfo still reports HT when you look at iwinfo.

And it's the same issue for the ipq4019, it's the exact same driver issue, in the same driver.

aanon4:
Thanks.
Same driver, different firmware blob which is why I was double checking. Same driver for the Ubiquiti AC devices too and yet another firmware blob, but as you've noted they work better. This is likely because the Ubiquiti devices use the standard Candelatech firmware shipped with OpenWRT, while we are forced to use the DD-WRT firmware for Mikrotik/TPLink otherwise we get no 10MHz support at all.

grozzie2:
2440 and 2441 look correct from here.

The folks in Courtenay will be happy, they have a bunch of mikrotik devices that have been unable to hold a connection to the mountain top units for some time.

PR #2379, opened 2025-08-15
Allow setting of LAN VLAN for single port devices.

aanon4:


cboling2:
I was excited to see this feature in the latest documentation. I installed the 2025-08-25 Babel nightly (since the feature hasn't been merged into main) and see that both WAN and LAN appear in the WebUI, and the correct default values show (1 for WAN, untagged LAN).

But BOTH fields (not just the new one) seem to have the same problem: changes made don't "stick" -- they don't trigger the "Pending changes:" banner, and if I force that to appear by changing another option at the same time, they don't contribute to the change count, and when I back in, the default values still show.

aanon4:
5 or above (others are used internally). I'll improve the docs and validation.

cboling2:
Hmmm, I do remember seeing that lower "5" limit a while back, now that you mention it.... Since it quietly ignored it instead of saying "Hey, dummy!", I didn't even consider the problem being me entering an illegal number.  Thanks, Tim & Steve!  Works as advertised w/ 5,6 instead of 3,4.

PR #2422, opened 2025-09-04
Improved log watch command

aanon4:


aanon4:
I type this sequence of command a lot while debugging.

PR #2440, opened 2025-09-09
Improve HTMODE selection (with thanks to https://github.com/grozzie2)

aanon4:


aanon4:
NOHT is only used for dd-wrt firmware radios when in mesh mode. Otherwise we act as "normal".

PR #2482, opened 2025-09-25
Rework the tunnel config message field names to match UI.

aanon4:


files/app/main/status/e/tunnels.ut
```diff
@@ -514,18 +513,21 @@ This value is used by default, but each tunnel may overide it.
                 }
                 const config = {};
                 txt.split("\n").forEach(line => {
-                    if (line.startsWith("Password: ")) {
-                        config.passwd = line.substring(10);
-                    }
-                    else if (line.startsWith("Wireguard key: ")) {
```
darrylhodgins, line 520: Consider retaining these for backward compatibility?

PR #2531, opened 2025-10-22
Initial support attempt for Cudy TR1200 (mt76x8)

uguraltinsoy:
This PR adds an initial device configuration for the Cudy TR1200 travel router, which uses the MT76x8 chipset (similar to OpenWRT One).

The firmware builds successfully with:

```make compile TARGET="cudy-tr1200" MAKE_ARGS="-j1 V=s"```

However, the following points are important:

- LAN and WAN ports require a custom network initialization script (`init-network.sh`) to work correctly.  
- **The script must be called from `/etc/rc.local` after installation**, so that it runs automatically on every boot.  
- DTDLink is not functioning.  
- 2.4 GHz Wi-Fi channels (-1, -2, -3, -4) are visible, but no mesh connection is established.  
- 5 GHz Wi-Fi LAN Hotspot mode works correctly.

The purpose of this PR is to **add initial support for the Cudy TR1200 device** and to contribute towards full device integration in AREDN.

Files included:
- `custom-added/cudy-tr1200.config`
- `custom-added/init-network.sh`

Any feedback, guidance, or corrections from the team are greatly appreciated.

### DIFF
--- a/files/etc/mesh-release
+++ b/files/etc/mesh-release
@@ -1,1 +1,1 @@
-TA2DMX-cudy-tr1200-support-12747679
+TA2DMX-cudy-tr1200-support-a7327fed
--- a/files/etc/radios.json
+++ b/files/etc/radios.json
@@ -1362,6 +1362,9 @@
   "pc": {
     "name": "PC"
   },
+  "cudy tr1200 v1": {
+    "name": "Cudy TR1200 v1.0"
+  },
   "unknown": {
     "name": "Unknown"
   }
--- a/patches/756-ramips-cudy-tr1200.patch
+++ b/patches/756-ramips-cudy-tr1200.patch
@@ -1,12 +1,15 @@
---- openwrt/target/linux/ramips/mt76x8/base-files/etc/board.d/02_network.orig	2025-10-26 02:42:07.124410392 +0300
-+++ openwrt/target/linux/ramips/mt76x8/base-files/etc/board.d/02_network	2025-10-26 02:42:45.414203991 +0300
-@@ -102,6 +102,10 @@
+--- openwrt/target/linux/ramips/mt76x8/base-files/etc/board.d/02_network.orig	2025-10-28 14:38:55.830014900 +0300
++++ openwrt/target/linux/ramips/mt76x8/base-files/etc/board.d/02_network	2025-10-28 14:39:16.060933303 +0300
+@@ -100,8 +100,11 @@
+ 		;;
+ 	cudy,m1200-v1|\
  	cudy,tr1200-v1)
- 		ucidef_add_switch "switch0" \
- 			"0:lan" "1:wan" "6@eth0"
-+		ucidef_set_interface_lan "eth0.1"
-+		ucidef_set_interface_wan "eth0.2"
-+		ucidef_set_interface "dtdlink" protocol "none"
+-		ucidef_add_switch "switch0" \
+-			"0:lan" "1:wan" "6@eth0"
++		ucidef_add_switch "switch0"
++		ucidef_add_switch_attr "switch0" "enable" "false"
++		ucidef_set_interfaces_lan_wan "eth0" "eth0.1"
++		ucidef_set_interface "dtdlink" device "eth0.2" protocol "static"
 +		ucidef_set_interface "wifi" device "wlan0" protocol "static"
  		;;
  	cudy,wr1000)

### CODE REVIEW

**Review by aanon4 — CHANGES_REQUESTED** [review]
There's a lot of problems here I'm afraid. This needs to integrate with the current code base in a way which will build with the current code base. Please take a look at how we do that. There are lots of examples in the patches/ directory which demonstrate how new devices are added. In particular look at how we configure the network. You dont need the 'init-network.sh' file if you integrate it like other devices.

**Comment by aanon4** [comment]
By the way, the OpenWRT One is ARM based, while this is using a much older MIPS core. The Cudy TR3000 is a closer comparison.

**Comment by uguraltinsoy** [comment]
> How old is this device? It only has 100Mbps ethernet ports.

I believe it was released in 2024. The Ethernet side isn't very good because it's cheap.

**Comment by uguraltinsoy** [comment]
> Please update this PR as we go along so, eventually, it'll be in a place where I can merge it. That's the final goal.
> 
> How are you testing firmware? Do you use these instructions: https://openwrt.org/toh/cudy/tr1200?#oem_easy_installation
> 
> Before you do try installing we need to redo the networking, but I'll wait until this PR is updated first.

I made a Pull Request for the project. What should I do now?

**Comment by aanon4** [comment]
Setting up the network correctly involves creating a patch which is applied to OpenWRT when it's built. You cannot just change the source code because that's now how OpenWRT is built and changed - it uses a base install onto which patches are then applied. It's messy but allows OpenWRT to upgrade their upstream dependencies easily.

So, you need to create a similar patch for this Cudy device which will be applied by the build system when you build your image. The cleanest example of how to do this is the file `patches/754-openwrt-one.patch` which looks like this:
```
--- a/target/linux/mediatek/filogic/base-files/etc/board.d/02_network
+++ b/target/linux/mediatek/filogic/base-files/etc/board.d/02_network
@@ -87,6 +87,8 @@
 	openembed,som7981|\
 	openwrt,one)
 		ucidef_set_interfaces_lan_wan eth1 eth0
+		ucidef_set_interface "dtdlink" device "eth1.2" protocol "static"
+		ucidef_set_interface "wifi" device "wlan0" protocol "static"
 		;;
 	dlink,aquila-pro-ai-m30-a1|\
 	dlink,aquila-pro-ai-m60-a1)
```
Each OpenWRT target has a network setup file with instruction similar to this which tell it how to configure things. The example above is for the OpenWRT One device, and the patch makes changes to this so it'll setup for AREDN. You will need to find the similar file for this Cudy device and write a similar patch to make that setup the network appropriately. This is what your network hack file was doing, but this is the official way OpenWRT, and so AREDN, does it.

aanon4:
Your patch probably isnt being applied if you check it. Make sure you have added it to `patches/series`

Also, I ordered one of these routers. It should turn up shortly so I can verify this as we go along.

uguraltinsoy:
I changed the patch number from 755 to 756 because another file was already using 755. Then, I added the file to `patches/series`, built the firmware, and updated the existing system using the new .bin file. However, the WAN port is not working  only the LAN port functions, and DTD is not working either.
Do VM devices support the DTD feature?

uguraltinsoy:
I tried two different patches, but I still couldn’t get the WAN to work. The LAN is functioning, but I think DTD might require an additional LAN port. This device only has one LAN and one WAN port.
Also, when scanning 2.4 GHz Wi-Fi, it scans all networks instead of just the AREDN ones.

aanon4:
How are you testing these? The PR I posted and yours look identical (bar the build bits which dont matter) and I tested mine on the device and it works fine.

aanon4:
Let's start with DTD which is on VLAN2. You say above you're connecting to a VM ... Proxmox or something else? How do you have the VLANs configured in the hypervisor and how are you physically routing them out of the VM to the device?

aanon4:
And you did support the project - none of this would have happened without your efforts. Thanks for everything.

uguraltinsoy:
> (1) The WAN port works fine, but it needs to have VLAN 1 configured on your switch. I dont know how to configured your switch. (2) Wifi scan will see everything. If -ve channels dont work it means the chipset doesnt support it and we'll need to exclude them from being selectable. Happy to do that it you confirm what does and doesnt work.

Of course, as I continue running tests, I’ll share all my feedback to help improve the project and assist with updates whenever possible.
I also have a different question — I’m not sure if it’s appropriate to ask here, but we don’t appear on the AREDN map when using the default SSIDs. Do you know why that might be? What should be done to fix this?
At the moment, we have around 12 RF devices active and are deploying about 40–50 more, but they don’t show up on the map with the default firmware. What should we do to make them visible?

aanon4:
The mapping works by scanning all the meshes which are connected together via "supernodes". You dont appear because you folk dont have one connected to the rest. You can find some information about it this here https://docs.arednmesh.org/en/latest/arednHow-toGuides/supernodes.html Let me know if you want to discuss hardware to use for this. I would love to make this happen.

uguraltinsoy:
> I dont really understand. If you are running AREDN how are you doing an upgrade in OpenWRT?

My mistake  sorry about that  I realized the problem was caused by not removing the check listed below. On initial setup, I can't directly flash the official Cudy TR1200 firmware (https://downloads.arednmesh.org/afs/www/?version=babel-20251102-45e6b6e9&target=ramips%2Fmt76x8&id=cudy_tr1200-v1) to my device as a SYSUPGRADE, because there is no compatible kernel version yet. For this reason I first install OpenWRT on the device and then install the AREDN firmware. Below I show step-by-step how I install it. 

1.  First of all, I'm coming to the place of Software in Cudy Software
Jrh5zFhWa3

2. When I try to install the SYSUPDATE file, I get this error, so I am installing OpenWRT.
zen_HpHYQGyZQc

3. I am installing the AREDN file as a system update in OpenWRT.
lNbmBZS8Hw

2BFbnOAGrW

aanon4:
I doubt very much we'd disable the DtD VLAN on a standard build. That doesnt stop you from trying it in a custom build though. However, lots of the software assume that's going to be there, so I dont even know if it's possible.

If you wanted to try without a configured DtD VLAN, and with LAN and WAN just being normal ports, you could try removing the part of the patch which changes the switch configuration. I dont know if that's sufficient for it to work but it's something you could experiment with.

uguraltinsoy:
Hello,
I was testing the Cudy TR1200. The DtD link is working, but the RF-side tests have failed. I performed the tests using two Cudy devices, and I’ve included the test results below. I believe there might be another issue causing this.

For 5 GHz, I tried the supported channels listed in the [datasheet](https://www.lcsc.com/datasheet/C45354768.pdf), but none of them worked. However, I noticed something interesting during the tests — it is able to transmit on channels -1, -2, and -3. Also, when the device is in LAN Hotspot Mode, my computer can detect its Wi-Fi signal.

| BAND | CHANNEL | BAND WIDTH | IS WORK |
|------|----------|-------------|----------|
| 2.4  | -4       | 20MHz       | NO       |
| 2.4  | -4       | 40MHz       | NO       |
| 2.4  | -3       | 20MHz       | NO       |
| 2.4  | -3       | 40MHz       | NO       |
| 2.4  | -2       | 20MHz       | NO       |
| 2.4  | -2       | 40MHz       | NO       |
| 2.4  | -1       | 20MHz       | NO       |
| 2.4  | -1       | 40MHz       | NO       |
| 2.4  | 1        | 20MHz       | YES      |
| 2.4  | 1        | 40MHz       | YES      |
| 2.4  | 13       | 20MHz       | YES      |
| 2.4  | 13       | 40MHz       | YES      |
| 2.4  | 14       | 20MHz       | NO       |
| 2.4  | 14       | 40MHz       | NO       |
| 5.8  | 104      | 20MHz       | NO       |
| 5.8  | 108      | 20MHz       | NO       |
| 5.8  | 108      | 40MHz       | NO       |
| 5.8  | 108      | 80MHz       | NO       |
| 5.8  | 112      | 20MHz       | NO       |
| 5.8  | 116      | 20MHz       | NO       |
| 5.8  | 140      | 20MHz       | NO       |
| 5.8  | 140      | 40MHz       | NO       |
| 5.8  | 140      | 80MHz       | NO       |
| 5.8  | 149      | 20MHz       | NO       |
| 5.8  | 149      | 40MHz       | NO       |
| 5.8  | 149      | 80MHz       | NO       |

### 2.4GHz -3 Channel
iMjk4JPPTS

### 2.4GHz -2 Channel
7rc7LbKtiI

aanon4:
Can we move this work on channels into a new Issue and out of this very closed PR? Thanks. Otherwise I'll keep not seeing these updates.

PR #2641, opened 2026-03-11
Add support for Orange Pi Zero 3 (Allwinner H618) - Sunxi Target

uguraltinsoy:
This PR introduces initial support for the Orange Pi Zero 3 (Allwinner H618) within the sunxi target. While internal Wi-Fi is currently non-functional, the device's significant CPU and RAM resources make it an ideal choice for Supernode operations and high-traffic mesh management.
Status / Tested Features:

- LAN: Internal 1Gbps Ethernet working.
- WAN: Fully functional via USB-to-Ethernet adapters (RTL8152 driver included and verified).
- DtD (Device-to-Device): Working via Ethernet/VLANs.
- System Stability: Stable under AREDN build environment (Sunxi-CortexA53).

Hardware Configuration:

Since the Orange Pi Zero 3 features a single on-board Ethernet port, the WAN interface has been successfully tested and implemented using external USB Ethernet adapters. This allows the device to operate as a full-featured mesh gateway.

Known Issues / In-Progress:

- Internal Wi-Fi: The Unisoc AW859A chipset is currently not operational in this target.
- UI Visibility: The "Radio" tab is hidden by default because no wlan0 is detected, but the device functions perfectly as a wired Supernode/Tunnel server.

Motivation:

To provide a high-performance, cost-effective hardware option. Its superior processing power compared to traditional MIPS-based routers allows it to handle multiple concurrent tunnels, video proxies, and heavy DtD link traffic without bottlenecking.

IMG_8580

Screenshot_20260312_010712


aanon4:
I can give you feedback on the PR if you'd like; but I don't know whether we'll ultimately accept it. We don't currently support builds for barebones boards. We've previously rejected builds for the Raspberry Pi. There will need to be an internal discussion about this.

That doesn't stop you supporting your own build of course and, regardless, I'm happy to help you get this working. Let me know what you'd like to do.

files/etc/mesh-release
```diff
@@ -1 +1 @@
-KN6PLV-babel-39df733f
+TA2DMX-main-6030f5e9
```
aanon4: This file should be removed from the PR. We dont change this (or try not to ... sometimes accidentally, but we're not suppose to).

aanon4:
> I have applied the improvements you mentioned. I hope I didn’t miss anything.
> 
> However, I didn’t fully understand the part about the Radio section.
> 
> When I check the board name, I see the following:
> 
> ```
> root@TA2DMX-O:~# cat /tmp/sysinfo/board_name
> xunlong,orangepi-zero3
> root@TA2DMX-O:~# cat /etc/board.d/02_network
> ```
> 
> Based on the name I see there, I’m not sure what exactly I should modify or how it should be implemented. Could you please clarify how this part should be handled?

The radios file uses the lowercase result from this command `/usr/local/bin/get_boardid` (if memory serves)

aanon4:
Perhaps - but they're still bare bone boards which is why inclusion in the default build was denied.
As I said, nothing stops you making these builds available to other people. We have others who do this. For example, K3PGM make his builds for older unsupported hardware available here https://downloads.k3pgm.radio/afs/www/

aanon4:
Closing - AREDN does not currently accept bare boards into the main tree.

Issue #2594, opened 2026-01-05
Wireguard Tunnel Issues Solved by Lowering MTU


dbsanders:
I host about 8 WireGuard tunnels. One of the tunnel clients is connecting to me from the T-Mobile network, which uses CGNAT. His node would only show up as a MAC address in the neighbors list, and never a hostname. In his mesh status page, he would never see any node other than his own. It's as if the tunnel was "almost" established (the key was correct etc.), but not completely set up with routing nor hostname resolution.

After reloading firmware and other troubleshooting steps, the fix was lowering the MTU on the WireGuard server end, just on his WireGuard interface (wgc9 in this case), from 1420 to 1280. Once I did this, the tunnel came up immediately and is fully functional. It's possible a higher number may work but didn't try.

In environments where a WireGuard peer connects over cellular, CGNAT, IPv6-encapsulated, or otherwise constrained internet paths (e.g., T-Mobile, LTE/5G, Starlink), the effective path MTU can be significantly lower than expected and Path MTU Discovery (PMTUD) often fails due to ICMP filtering. It appeared that packets were getting dropped due to this condition.

"ip link set dev wgc9 mtu 1280" did the trick, but of course this will be 1420 again after a reboot.

**Describe the solution you'd like**
I would love to see either
1. A field in the tunnel setup (both server and client as it could affect both directions I believe), where one could set or choose a lower MTU value than the 1420 default. 
2. A checkbox that would enable "Starlink/CGNAT" friendly mode so one wouldn't need to choose a number. The docs/help would have to make it clear that this checkbox doesn't enable a WireGuard server to be hosted on Starlink/CGNAT as that's not possible - only a client.

**Describe alternatives you've considered**
Delayed rc/init script to set the correct value? Not great.

**Additional context**

[supportdata-K6BAT-IDEC-x86-QTH-2026-1-4-21-17.tar.gz](https://github.com/user-attachments/files/24427530/supportdata-K6BAT-IDEC-x86-QTH-2026-1-4-21-17.tar.gz)

dbsanders:
> I also ran into this with a remote node we are trying to get going (a tower sight that will eventually be a hub, but we need to link it into the network while the crew is trying to aim the beams and get it connected).
> 
> Is there a work around that can be manually applied so the `ip link set dev wgc## mtu 1280` persists between reboots?

[...]

PR #2665, opened 2026-04-04
Auto-adjust WireGuard MTU on interface bring-up to handle low-MTU uplinks (#2594)

thegeekoftheworld:
## Summary

This change adds automatic WireGuard MTU tuning during `iface` hotplug `ifup` events to help address the tunnel behavior described in #2594.

Issue #2594 describes WireGuard tunnels that appear to come up only partially on constrained paths such as T-Mobile/CGNAT, where lowering the WireGuard interface MTU resolved the problem immediately. The reported workaround was manually setting the affected WireGuard interface from `1420` down to `1280`, but that setting did not persist across reboot. 

## What this change does

When an `ifup` hotplug event occurs for either:

- `wan`, or
- a WireGuard interface matching `wgs[0-9]*`

the script will:

1. Wait 5 seconds to allow the WAN/path to stabilize.
2. Probe outbound path MTU by pinging `8.8.8.8`, starting at `1500`.
3. Step downward in increments of `20` until a working size is found, with a floor of `1000`.
4. Convert the discovered working WAN MTU into a WireGuard MTU by subtracting `80` bytes.
5. Apply that computed MTU to all WireGuard interfaces matching `wgs[0-9]+`.
6. Log the detected uplink ASN/organization when the discovered path MTU is lower than `1500`.

## Why

The issue report notes that constrained networks such as T-Mobile, CGNAT, IPv6-encapsulated paths, LTE/5G, and similar environments can have a lower effective path MTU than expected, and that PMTUD may fail when ICMP messages are filtered. In that report, lowering the MTU on the WireGuard interface resolved the tunnel problem. 

This implementation automates that adjustment at interface bring-up so the system can adapt after reboot or uplink changes without requiring a manual `ip link set dev ... mtu ...` workaround.

## Behavior notes

- This does **not** add a user-facing MTU field in the UI.
- This does **not** currently tune MTU per peer; it applies the computed WireGuard MTU to all `wgs*` interfaces on the node.
- The script only emits a log entry when the detected path MTU is below the normal `1500` value.
- The purpose is to make the workaround from #2594 persistent and automatic during interface activation.

## Related issue

Fixes / addresses #2594

### DIFF
--- a/files/etc/hotplug.d/iface/15-fixwgmtu
+++ b/files/etc/hotplug.d/iface/15-fixwgmtu
@@ -48,22 +48,24 @@ find_good_mtu() {
     high="$START_MTU"
     best="$MIN_OUTER_MTU"
 
+    # Phase 1: coarse binary search on 40-byte boundaries
     while [ "$low" -le "$high" ]; do
         mid=$(( (low + high) / 2 ))
-        mid=$(( mid - (mid % 20) ))
+        mid=$(( mid - (mid % 40) ))
 
         [ "$mid" -lt "$MIN_OUTER_MTU" ] && mid="$MIN_OUTER_MTU"
 
         if test_ping_mtu "$target" "$mid"; then
             best="$mid"
-            low=$((mid + 20))
+            low=$((mid + 40))
         else
-            high=$((mid - 20))
+            high=$((mid - 40))
         fi
     done
 
-    start=$((best - 30))
-    end=$((best + 30))
+    # Phase 2: refine around coarse result on 10-byte boundaries
+    start=$((best - 40))
+    end=$((best + 40))
 
     [ "$start" -lt "$MIN_OUTER_MTU" ] && start="$MIN_OUTER_MTU"
     [ "$end" -gt "$START_MTU" ] && end="$START_MTU"
@@ -87,7 +89,6 @@ find_good_mtu() {
 get_real_dev() {
     iface="$1"
 
-    # Hardcode common OpenWrt/AREDN logical WAN bridge
     if [ "$iface" = "wan" ] && [ -d /sys/class/net/br-wan ]; then
         echo "br-wan"
         return 0
--- a/files/etc/hotplug.d/iface/15-fixwgmtu
+++ b/files/etc/hotplug.d/iface/15-fixwgmtu
@@ -28,7 +28,6 @@ test_normal_ping() {
     ping -c 1 -W 3 "$target" >/dev/null 2>&1
 }
 
-# Test with a given MTU-sized packet using DF
 test_ping_mtu() {
     target="$1"
     mtu="$2"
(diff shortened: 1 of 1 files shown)


files/etc/hotplug.d/iface/15-fixwgmtu
```diff
@@ -0,0 +1,253 @@
...
+    fi
+
+    dev="$(ifstatus "$iface" 2>/dev/null | sed -n 's/.*"l3_device"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' | head -n 1)"
+    [ -n "$dev" ] && [ -d "/sys/class/net/$dev" ] && { echo "$dev"; return 0; }
+
+    dev="$(ifstatus "$iface" 2>/dev/null | sed -n 's/.*"device"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' | head -n 1)"
+    [ -n "$dev" ] && [ -d "/sys/class/net/$dev" ] && { echo "$dev"; return 0; }
+
+    return 1
+}
```
aanon4: Walk me though why this is so complex when hunting down the real interface name. Also, does $DEVICE help you here as it will know the device name.

Take a look here - https://openwrt.org/docs/guide-user/base-system/hotplug

files/etc/hotplug.d/iface/15-fixwgmtu
```diff
@@ -0,0 +1,253 @@
...
+        echo "$host"
+        return 0
+    fi
+
+    return 1
+}
+
+handle_wan() {
+    iface="$1"
+    target="$GOOGLE_TARGET"
```
aanon4: What happens then GOOGLE_TARGET is unreachable for some reason? The interface can come up without there being a route (think static IP).

files/etc/hotplug.d/iface/15-fixwgmtu
```diff
@@ -0,0 +1,253 @@
...
+        | sed -n "s/.*endpoint_host='\([^']*\)'.*/\1/p" \
+        | head -n 1)"
+
+    if [ -n "$host" ]; then
+        echo "$host"
+        return 0
+    fi
+
+    return 1
+}
```
aanon4: Why the need to interrogate the same uci file two different ways?

files/etc/hotplug.d/iface/15-fixwgmtu
```diff
@@ -0,0 +1,253 @@
...
+            log_msg "$iface up, endpoint $endpoint_host did not respond to ping, failed to set $iface MTU to safe minimum $wg_mtu"
+        fi
+    fi
+
+    return 0
+}
+
+if [ "$ACTION" = "ifup" ]; then
+    case "$INTERFACE" in
+        wan|wgs[0-9]*)
```
aanon4: Why no wgc devices?

aanon4 (commented):
I apologies if you think I'm being harsh here, but remember that this is the network code that every AREDN device will run all the time. It always needs to work, it cannot accidentally fail and break a node, and there are countless weird setups out there.

And I haven't even tested it yet - sorry. At some point this will need to be deployed on various different tunnel servers and supernodes for testing.

I will add that i do want to see this merged, but the nature of what's being changed means it will take a bit of work yet.

aanon4:
If you want a less problematic path forward for this, you should make these changes into a package so people who need it can install it. That way I dont have to worry about its effect on everyone.

PR #2669, opened 2026-04-06
UI: Sort wireguard tunnels by type and numeric index

uguraltinsoy:
Summary:
This PR improves the Wireguard tunnels list by introducing a more consistent and user-friendly sorting mechanism.

Problem:
Previously, tunnels were displayed based on creation/UCI order, which made management difficult—especially with many tunnels. Alphabetical sorting also caused incorrect ordering (e.g., client_11 appearing before client_2).

Solution:
A natural sorting algorithm was added to tunnels.ut:

server_N entries are prioritized and shown first, followed by client_N entries
Numeric indices are sorted numerically instead of lexicographically (e.g., client_2 < client_10)

Result:

More predictable and organized list
Improved usability for nodes with many tunnels
Clearer separation between server and client entries

Test:

Navigate to /a/status/e/tunnels
Verify that server entries appear before client entries
Ensure entries with the same prefix are sorted in ascending numeric order

### DIFF
--- a/files/app/main/status/e/tunnels.ut
+++ b/files/app/main/status/e/tunnels.ut
@@ -176,7 +176,6 @@ uciMesh.foreach("wireguard", "client", a => {
     delete available[a.clientip];
     servercount++;
 });
-
 sort(tunnels, (a, b) => {
     const ma = match(a.index, /^([a-z]+)_(\d+)$/);
     const mb = match(b.index, /^([a-z]+)_(\d+)$/);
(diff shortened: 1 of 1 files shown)


aanon4:
If we were to sort these, surely they should be sorted by something visually obvious to the user rather than an internal name?

uguraltinsoy:
> If we were to sort these, surely they should be sorted by something visually obvious to the user rather than an internal name?

During my usage, I found it more logical and accurate to sort by port numbers rather than names. Because sometimes when tunnels stop working, they get deleted and recreated, which can cause confusion. What are your thoughts on this?

resim

aanon4:
My thought process went - on the main status page the Neighborhood Nodes are listed in alphabetical order, so given we use this ordering there, perhaps it makes sense to use it here.

PR #2698, opened 2026-05-16
Crude way to set the MTU of all wireguard tunnels.

aanon4:
Provide a simple, if crude, way to override the mtu of all wireguard tunnels.

Edit the file `/etc/config.mesh/wireguard` and find the section `config network`. Add an option setting your new mtu value, for example `option mtu '1400'`. Reboot the node for this to take effect.

On my node, the whole `config network` section might look like this:
```
config network 
        option begin '172.31.64.76' 
        option dns 'aredn.xojs.org' 
        option mtu '1400'
```


AI7BQ:
Should this workaround be implemented at both the tunnel client and tunnel server endpoints?

aanon4:
You care about the minimum MTU along the path, so yes - you need the same mtu setting at both ends.

PR #2702, opened 2026-05-17
Bump dnsmasq to 2.92rel2

aanon4:


aanon4:
Various reported DNS problems and crashes. OpenWRT is a couple of versions behind and there are known memory fixes in the latest release.

PR #2704, opened 2026-05-20
Gl a1300

md-thomas:
I'd like to request a PR.  I have added support the the GL-A1300 GLiNet Travel Router.


config.mk
```diff
@@ -1,7 +1,7 @@
 # default parameters for Makefile
 
 # What version will show in the AREDN® firmware 'Node Status" UI page?
-PRIVATE_BUILD_VERSION=NoCall
+PRIVATE_BUILD_VERSION=K0MDT
 
 # build options:  -j# for (# of cores +1) on build machine,  V=s for verbose output
 # https://wiki.openwrt.org/doc/howto/build#make_options  (archive)
```
aanon4, line 7: This file shouldn't be checked in.

files/etc/mesh-release
```diff
@@ -1 +1 @@
-KN6PLV-babel-39df733f
```
aanon4, line 1: This file should not be checked in.

aanon4 (changes requested):
I've noted a couple of files which should not be checked in.
You should also add yourself to the CONTRIBUTORS file if you're not already.
Tell me about testing and if there are known issues?

files/usr/share/ucode/aredn/hardware.uc
```diff
@@ -859,6 +859,8 @@ export function getEthernetPorts()
         case "cudy,wr3000p-v1":
         case "cudy,wr3000s-v1":
             return default5PortLayout;
+        case "glinet,gl-a1300":
+            return default3PortLayout;
```
aanon4: Let's just let this fall-through to the same default3PortLayout as it's the same port layout from the same manufacturer. I'm trying to group these when I can.

aanon4 (changes requested):

md-thomas:
OK, I think I have those 3 changes in.
I compiled and installed the firmware on my GL-A1300.
It's been running for several days now.

You can see it here:  K0MDT-WCF-QTH-3

Initially I thought the 2 LAN ports were both LAN, but after digging in
some, I discovered that one port is the
DTE port. That is super handy. I have a phone connected to the LAN port and
its receiving its IP address from
the node.

So far, everything looks good.  I have had the Radio turned on and off.

Let me know what else you need.

-MikeT

SUPPORTED_DEVICES.md
```diff
@@ -141,6 +141,7 @@ Model | SKUs | Band | Target | Subtarget | Image | RAM | Stability | Status
 Shadow (16MB NOR) | GL-AR300M16 <br> GL-AR300M16-Ext | 2 | ath79 | generic | glinet_gl-ar300m16 | 64MB | stable | released
 Shadow (128MB NAND) | GL-AR300M <br> GL-AR300M-Ext | 2 | ath79 | nand | gl-ar300m | 64MB | stable | released
 Mudi | GL-E750 | 2 & 5 | ath79 | nand | gl-e750 | 128MB | stable | stable
+Slate Plus | GL-A1300 | 2 & 5 | ipq40xx | generic | gl-a1300 | 256MB | stable | released (6)
```
aanon4: This should be marked as "nightly" and not release as it's not yet in a release.
Also, the (6) indicated that -ve channels dont work, but I also noticed that K0MDT-WCF-QTH-3 is currently configured to use -2. Curious it that actually works and the (6) is in error.

aanon4 (changes requested):

md-thomas:
- The (6) was my fault, I didn't understand that correctly.
- Understand the nightly, now.

I'll turn on a radio on another node, and test that before I push again.

md-thomas:
- (6) seems to be correct.  The node does not appear to work on negative
2GHz frequencies.
I'm glad you mentioned that.

I'll do some more testing, to see if I missed anything else.
Do you test each frequency or just spot check across the band?

aanon4:
Thanks for checking. I was expecting the -ve channels not to work, but as you had configured it I thought I might as well ask because you never know.

We mostly spot check, especially when a device is very much like another we're more familiar with.

aanon4:
There's one minor change still pending. More a stylistic thing but if you could resolve it I can merge this - thanks for your work here - it's very much appreciated.

md-thomas:
What is the last change that is needed?
And, is there a way to disable those negative frequencies per node type, or
that is baked into the UI?

aanon4:
If you look back through this conversation you should see one open item still - it's a tweak related to the selection of the 3port layout.

Disabling channels - I added something to do this last night after our discussion - look at this PR to help you add the exclusions for this radio - https://github.com/aredn/aredn/pull/2708/changes - specifically the addition to the radios.json file.

md-thomas:
I'm following you now. I like the exclusion of the channels, that should
prevent issues with people asking why its not working.
That was a nice addition.  I also found what you were talking about for the
stylistic update.  I see how that is also better.
Thank you for being patient with me.  I had some issues with the rebase,
but I finally got it to take the updates.

aanon4:
It looks like you picked up a couple of unnecessary changes during your rebase which need to be removed.

hardware.uc
790-morsemicro-support.patch

hardware.uc has your valid changes in it - it just has some other stuff too. the 790-X file shouldnt be there at all.

md-thomas:
in hardware.uc I'm changing this
case "glinet,gl-a1300":
            return default3PortLayout;
case "glinet,gl-b1300":
           return default3PortLayout;

to this

case "glinet,gl-a1300":
case "glinet,gl-b1300":
            return default3PortLayout;

Is that not correct?

I don't see any updates to 790-omrsemicro-support.patch
when I git rebase main
I get merge conflicts for
  Auto-merging SUPPORTED_DEVICES.md
  CONFLICT (content): Merge conflict in SUPPORTED_DEVICES.md
  Auto-merging files/etc/radios.json
  CONFLICT (content): Merge conflict in files/etc/radios.json
  Auto-merging files/usr/share/ucode/aredn/hardware.uc
  CONFLICT (content): Merge conflict in
files/usr/share/ucode/aredn/hardware.uc

Those are the 3 files I'm updating.

aanon4:
If you look here - https://github.com/aredn/aredn/pull/2704/changes - it's give a new tree view of what changes and files are in your PR.

aanon4:
Isn't git great :-)

So what you've done now is rather then deleting the files from your PR, you have deleted the files from the git repo instead (or would do if this was pushed). Probably not what you wanted to do.

md-thomas:
Yes, sometimes it's great :), sometimes it's frustrating. :(
I saw that, I undid that commit, and rm --cached this time.

md-thomas:
I did step in it didn't I :)

md-thomas:
hardware.ui is back, and I don't see 790-morsemicro-support.patch in the
repo atm

md-thomas:
now that hardware.ui is back, do I need to git rm --cached hardware.ui  to
remove it from the PR?

md-thomas:
hardware.uc I meant to type

aanon4:
Unfortunately there's still odd bits in the PR which shouldn't be there.

At this point I'm going to suggest you create a new fresh branch, make sure its up to date with the current aredn sources (which were updated a couple of hours ago), apply your changes to this branch by hand, and then submit that branch as a new PR. It'll just be easier at this point.

md-thomas:
I agree.  I'll try and work on it this afternoon.

md-thomas:
Just an update.
After starting over, and building.
The build seemed fine.  But that firmware won't run.
The node comes up with 192.168.1.1 I can ping it, but the web page never
loads.
The first firmware build still works.
I noticed some new updates on main, so I'm trying again.
I'll keep building and testing, it's good practice anyway.
I'm not sure if there is a way to get into the device to see what might be
wrong.
I don't think ssh worked either, but I tried so many times, it might.  I'll
try that again.

aanon4:
If the new branch is available on your github I can take a look.

md-thomas:
Yep, I pushed it.  Its be re-sync'd with aredn/main, and I'm building new
again.
I put the config.mk and files/etc/mesh-release in my .gitignore.

md-thomas:
I am able to ssh in.
Not sure what to look at... maybe /var/log...

md-thomas:
The problem is def. in the hardware.uc file, trying to figure out where.
maybe those new lines...

md-thomas:
yep, found it.

md-thomas:
it was actually in the radios.json, that hardware.uc was reading.
I was missing a '}'

md-thomas:
Yep that was it. Its loading the new FW now.
I figured this out by sshing in and running logread | tail -50

I'll test it, and if it looks good, I'll submit a PR.

md-thomas:
I think there is a problem with 2GHz Channel 14.
I have this node, and one other that I'm testing with.
All the other channels I tested both nodes could see each other.
On channel 14, they cannot.

md-thomas:
Yeah, so channels 1 to 13 work fine, but channel 14 does not.

md-thomas:
I found this on the Internet.
The GL.iNet GL-A1300's 2.4 GHz radio will not work on Channel 14 because
its use is legally prohibited outside of Japan. By default, the firmware
enforces FCC/regional power and channel rules, entirely locking out
"illegal" channels to maintain regulatory compliance. [1
<https://forum.gl-inet.com/t/mt6000-wrong-hardware-country-code/53466>, 2
<https://www.acrylicwifi.com/en/blog/forbidden-wifi-channels-by-country/>]
If you absolutely need to test or connect via Channel 14 for specialized
legacy hardware (like 802.11b devices), you will need to bypass regional
limits by accessing the router via SSH and forcing the regulatory country
code to Japan ("JP").

PR #2725, opened 2026-05-27
Enable QCA955x on-chip WMAC radio on Ubiquiti XC boards

mathisono:
Enable the AHB-attached 802.11n radio present in the QCA9558/QCA9560 SoC on the Ubiquiti XC board family (PowerBeam 5AC 500, Rocket 5AC Lite, NanoBeam AC XC). This radio can serve as a dedicated spectral scanner for RF interference detection without disrupting the main ath10k mesh radio.

The WMAC hardware is present in the SoC and has valid AR9300 calibration data in the ART partition at offset 0x5000. The existing cal_art_5000 nvmem cell is shared - both ath10k and ath9k read different fields from the same caldata block. The ath9k driver and platform driver are already loaded in AREDN firmware; this change enables device tree probing so ath9k binds to the hardware and exposes phy1.

The patch creates a shared include (qca955x_ubnt_xc-wmac.dtsi) so all three XC boards get the change. The pattern follows 11+ existing upstream OpenWrt QCA955x devices that enable the WMAC (Archer C7 v1, Aruba AP-115, LibreRouter v1, etc.).

Tested: patch applies cleanly against OpenWrt v25.12.4 (AREDN base). Hardware verification pending - requires firmware build and flash to bench device.

PR #2730, opened 2026-05-30
Provide hooks to enable the spectrum radio on specific Ubiquiti AC radios.

aanon4:
This change will enable the spectrum radio on the Rocket 5AC Lite and the PowerBeam 5AC 500 which are the designated test devices. However this change does *not* include the calibration data to bring these radios up, so they will not initialize. The job of providing this data is left to the app which operates the spectrum analyzer.

See the original proposal here https://github.com/aredn/aredn/pull/2725

### DIFF
--- a/files/etc/radios.json
+++ b/files/etc/radios.json
@@ -851,9 +851,14 @@
   },
   "0xe1f5": {
     "name": "Ubiquiti Rocket 5AC Lite",
-    "maxpower": 23,
-    "pwroffset": 4,
-    "antenna": "external",
+    "wlan0": {
+      "maxpower": 23,
+      "pwroffset": 4,
+      "antenna": "external"
+    },
+    "wlan1": {
+      "disabled": true
+    },
     "features": [ "xlink" ]
   },
   "0xe202": {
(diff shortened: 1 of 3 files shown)


mathisono:
wanted to note where my downstream work sits relative to this PR.

radios.json — now aligned ✅
I've adopted the wlan0/wlan1 nesting structure from this PR for the Rocket 5AC Lite (0xe1f5) and PowerBeam 5AC 500 (0xe3d5). My local build now matches these changes exactly.

DTS patch — different approach, same goal ⚠️
I submitted a separate PR earlier ("Enable QCA955x on-chip WMAC radio on Ubiquiti XC boards") that takes a slightly different approach to enabling the wmac:

Shared .dtsi include (qca955x_ubnt_xc-wmac.dtsi) instead of inline &wmac blocks in each board DTS — keeps things DRY.
Real caldata wired up via nvmem-cells = <&cal_art_5000> instead of qca,no-eeprom. The AR9300 compressed EEPROM at ART offset 0x5000 is valid for ath9k on these boards (confirmed on bench hardware), so the radio can fully initialize and produce spectral FFT data immediately without the app needing to inject calibration.
Covers NanoBeam AC XC in addition to the Rocket 5AC Lite and PBE-5AC-500, since it's the same QCA9558 SoC with the same disabled wmac.
Happy to reconcile the two DTS approaches however makes sense for upstream. The radios.json piece is the important structural alignment and that's done on my end.

mathisono:
https://github.com/mathisono/aredn/commit/3e45ed0b289a91d09dc4638dff6b0db375306c00

Issue #2746, opened 2026-06-04
ALFA Network Tube-AHM Channel Display is wrong.


KenHorse:
### Describe the bug
The channel displays between the actual selected channel and what the UI says is the selected channel, differ.

Running 20260603-6aaf7a5f

### Expected behavior
They should match!

[supportdata-AH6LE-PARRETT-HALOW-2026-6-3-17-16.tar.gz](https://github.com/user-attachments/files/28575137/supportdata-AH6LE-PARRETT-HALOW-2026-6-3-17-16.tar.gz)

### Support data
Please add the support data for all nodes mentioned in this bug report (**Required**)

### Screenshots
<img width="652" height="336" alt="Image" src="https://github.com/user-attachments/assets/9df2d09c-31de-4384-b09d-e266292445bb" />

<img width="520" height="425" alt="Image" src="https://github.com/user-attachments/assets/2c12a845-5e00-4e9a-8e8c-1f9cb803e36c" />

aanon4:
When I can select channel 8 or 12 at 4 MHz, the main display always shows the correct channel.

I need steps to reproduce this - I cannot right now.

KenHorse:
Channel 12 shows on the landing page but doesn't even show as an option in the dropdown from the popup. 
1) Click on Radio window, popup opens.
2) Channel twelve is not even an option in the dropdown, select 8
3) Note SSID is wrong too.
4) Click DONE, Commit and reboot. 
5) Radio display shows Channel 12

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/f4ce0eb8-dc55-4df1-90ef-beb932548ce7" />

KenHorse:
I did not mean to close this... sheesh......

aanon4:
What happens with other channel selections?

KenHorse:
I'll try some other channel selections from the dropdown. It takes these suckers 5 minutes to reboot!

aanon4:
I followed your steps, and it still says channel 8 for me.

Was this previous set to channel 12 on some other MHz setting?

aanon4:
> It takes these suckers 5 minutes to reboot!

Yes, they really cheaped out when they chose a processor for this thing.

KenHorse:
I never set it to Ch 12. In fact, when AREDN was first installed, it defaulted to Ch 5. I've done 2 and they both (mis)behave the same way.

KenHorse:
[...]

PR #2750, opened 2026-06-09
Fix bug when channel didnt change when selection new bandwidth sometimes

aanon4:


aanon4:
https://github.com/aredn/aredn/issues/2746

PR #2777, opened 2026-06-18
GL-E750 V2 not currently supported.

aanon4:


aanon4:
https://forum.openwrt.org/t/is-there-already-support-for-gl-inet-gl-e750-v2/197542

PR #2805, opened 2026-07-17
Major network changes

aanon4:
* Radios \- It is now possible to select all the radio modes except WAN on multiple radios. The UI for this hasn’t changed.  
* XLinks \- you can no longer specify the IP addresses or CIDR for xlinks. These addresses are all generated internally. The UI has changed to remove these fields. Old xlink endpoints will still work with new ones.  
* Tunnels \- you no longer need to specify the IP addresses. These are generated internally. The old tunnels will continue to work, but the new configurations are not compatible with older firmware which still expects the IP addresses.  
  * Clients \- The Network:Port field has been removed.  
  * Servers \- The Network:Port field has been removed, but a new Port field has been added (unfortunately we still need a different port per tunnel) next to the Remote Node Name field.

  When entering a tunnel (you can still copy-and-paste as before) the port number now goes on the end of the server name \- e.g. my-tunnel-server.com:5525

* Under Advanced Options:  
  * The Tunnel Server Network has been removed as we no longer need it.  
  * A new Default Tunnel MTU option has been added to let people reduce the MTU for tunnels. This was previously available but not in the UI.  
* Internal Services  
  * WAN Monitor \- A new WAN monitor has been added (UI changes on the main pages as well as the Internal Services popup) which makes sure the WAN is working and stops advertising it to the Mesh (if the user has this enabled) if not. This was a feature built-in to OLSR but not Babel.  
  * Under Advanced Options:  
    * Added restart for LQM.  
* Internal changes to help reduce the need to reboot the node when changing stuff, especially radio settings.


K1RKS:
So great to see the network improvements. Especially the MTU which has been a thorn for two tunnel clients. Well done team!

PR #2872, opened 2026-08-16
add total ram

ab7pa:
What would you think of the idea to add "total ram" to the health display?
Memory varies a lot across the supported devices, and I often forget how much ram a node has available.
**This just prints the estimated total ram as an integer at the end of the list.**

Screenshot 2026-08-16 at 1 54 50 PM

### DIFF
--- a/files/app/partial/health.ut
+++ b/files/app/partial/health.ut
@@ -43,8 +43,9 @@
         uptime = "1 day, " + uptime;
     }
     const ld = split(fs.readfile("/proc/loadavg"), " ");
-    const ram = int(match(fs.readfile("/proc/meminfo"), /MemFree: +(\d+) kB/)[1]) / 1000.0;
-    const totram = int(match(fs.readfile("/proc/meminfo"), /MemTotal: +(\d+) kB/)[1]) / 1000;
+    const meminfo = fs.readfile("/proc/meminfo");
+    const ram = int(match(meminfo, /MemFree: +(\d+) kB/)[1]) / 1000.0;
+    const totram = int(match(meminfo, /MemTotal: +(\d+) kB/)[1]) / 1000;
     const f = fs.popen("exec /bin/df /");
     let flash = "-";
     if (f) {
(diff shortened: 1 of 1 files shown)


Orv:
This private-user-images.githubusercontent.com page can’t be found

:-(

files/app/partial/health.ut
```diff
@@ -44,6 +44,7 @@
     }
     const ld = split(fs.readfile("/proc/loadavg"), " ");
     const ram = int(match(fs.readfile("/proc/meminfo"), /MemFree: +(\d+) kB/)[1]) / 1000.0;
+    const totram = int(match(fs.readfile("/proc/meminfo"), /MemTotal: +(\d+) kB/)[1]) / 1000;
```
aanon4: Let's not read the data from the file system twice - read once, match twice.

aanon4 (changes requested):

files/app/partial/health.ut
ab7pa: Good catch!  Thanks, @aanon4

files/app/partial/health.ut
```diff
@@ -43,7 +43,9 @@
         uptime = "1 day, " + uptime;
     }
     const ld = split(fs.readfile("/proc/loadavg"), " ");
-    const ram = int(match(fs.readfile("/proc/meminfo"), /MemFree: +(\d+) kB/)[1]) / 1000.0;
+    const meminfo = fs.readfile("/proc/meminfo");
+    const ram = int(match(meminfo, /MemFree: +(\d+) kB/)[1]) / 1000.0;
+    const totram = int(match(meminfo, /MemTotal: +(\d+) kB/)[1]) / 1000;
```
aanon4, line 48: You need to divide by 1000.0 rather than 1000 otherwise the value is truncated to an integer. Unless that was the goal?

aanon4 (changes requested):

ab7pa:
The goal was to display just the integer value.

PR #2874, opened 2026-08-19
Allow babel to run over the LAN network.

aanon4:


KenHorse:
Is to replace xlinks?

aanon4:
No

